PR #2, opened 2018-10-08
Simple ZMON collector implementation

mikkeloscar:
This adds a simple collector for getting metrics from ZMON checks. It's mostly a proof of concept at this point but can be used for "simple" metrics.

## TODO

* [ ] Add tests
* [ ] Improve docs
* [ ] Add support for aggregations


Jan-M:
You should look at the endpoint used for the entities lookup, there is two options, one is powered by the PostgreSQL database and one is using the Scheduler that keeps all entities in memory to serve the UI entity search (when editing filters).

The database also powers all the agents every 60 seconds and distributed schedules and the jsonb query uses a proper index, but maybe the Scheduler endpoint may still be quicker.

In general I wonder the usefulness of resolving entities first over simply querying kairosdb (with version and application, should return the data one might look for).

mikkeloscar:
> You should look at the endpoint used for the entities lookup, there is two options, one is powered by the PostgreSQL database and one is using the Scheduler that keeps all entities in memory to serve the UI entity search (when editing filters).
>
> The database also powers all the agents every 60 seconds and distributed schedules and the jsonb query uses a proper index, but maybe the Scheduler endpoint may still be quicker.

This is running as a background job caching the results, so the speed is not so important. If there's one endpoint we should prefer to hit over another one then we can do that (I'll ask eagleeye).

> In general I wonder the usefulness of resolving entities first over simply querying kairosdb (with version and application, should return the data one might look for).

This is just to make it more flexible and support more than the default tags from kariosdb. Making an extra call is not an issue for this component, if it hurts ZMON we could try to avoid it, otherwise I wouldn't worry about it. I always find the disconnect between entity filter in ZMON alerts and the limited tags in Grafana annoying. The entity filter is much more powerful.

Jan-M:
I am with you on the general usefulness of the entity filter and meta data. So then this is def +1. The caching works how? How are new pods then discovered for example?

One should just be sure that the long list of explicit entity IDs to KairosDB works as expected. (Assuming e.g. skipper at 50+ entities).

mikkeloscar:
> The caching works how? How are new pods then discovered for example?

It works like this:

every minute (default) the collector (what this PR implements) goes to ZMON and gets the entities and then makes the kariosdb query. The result of the query is stored in-memory of this component and is what is served to the Horizontal Pod Autoscaler when it queries for metrics.

I'm not sure if the current intervals really makes sense, this is something still to be tested.

> One should just be sure that the long list of explicit entity IDs to KairosDB works as expected. (Assuming e.g. skipper at 50+ entities).

True, there are still things like this to test before this is ready. Thanks a lot for the feedback!

Jan-M:
Ok, that sounds right, thought you cache the entities.

mikkeloscar:
@mohabusama convinced me that we should support kariosDB tags instead of entities. The reason for this is that despite it being slightly less flexible it will allow a 1 to 1 mapping to queries created in Grafana dashboards making it much easier to debug in case a query does not work.

mikkeloscar:
@arjunrn updated (AND tested) this PTAL :)

mikkeloscar:
:+1:

PR #15, opened 2018-11-21
Skipper metrics collection for resource types

arjunrn:
Signed-off-by: Arjun Naik <arjun.rn@gmail.com>

The adapter attempts to collect metrics for `Resource` type and then records warnings in the HPAs. This should prevent that.


mikkeloscar:
:+1:

PR #22, opened 2018-12-15
Update aws-sdk-go version to be compatible with kube-aws-iam-controller

mikkeloscar:
This makes the metrics-adapter compatible with https://github.com/mikkeloscar/kube-aws-iam-controller

See this for reference on configuration: https://github.com/mikkeloscar/kube-aws-iam-controller/blob/master/docs/sdk-configuration.md#golang-aws-sdk


szuecs:
:+1:

PR #26, opened 2019-01-09
Weighted traffic metrics for request-per-second

arjunrn:
# Weighted traffic metrics for request-per-second

## Description
The ingress metrics were not working correctly when 2 HPAs were specifying the same ingress because of the calculation done based on the number of pods. But now `targetAverageValue` can be specified in the HPAs and this readjustment is not needed so it has been removed. Also the 1.12 API introduced support for label selectors on the metrics, which has been used to weight the returned metric based on the stackset and backend weights in the ingress.

## Types of Changes
- Feature: Weighting based on stackset and backend weights
- Fix: Removed adjustment of ingress weights

## Review
- [x] Tests
- [ ] Documentation


pkg/collector/collector.go
```diff
@@ -148,13 +149,13 @@ func (c *CollectorFactory) NewCollector(hpa *autoscalingv2beta1.HorizontalPodAut
 	return nil, fmt.Errorf("no plugin found for %s", config.MetricTypeName)
 }
 
-func getObjectReference(hpa *autoscalingv2beta1.HorizontalPodAutoscaler, metricName string) (custom_metrics.ObjectReference, error) {
+func getObjectReference(hpa *autoscalingv2beta2.HorizontalPodAutoscaler, metricName string) (custom_metrics.ObjectReference, error) {
 	for _, metric := range hpa.Spec.Metrics {
-		if metric.Type == autoscalingv2beta1.ObjectMetricSourceType && metric.Object.MetricName == metricName {
+		if metric.Type == autoscalingv2beta2.ObjectMetricSourceType && metric.Object.Metric.Name == metricName {
```
mikkeloscar: I wonder if this "just works" or if everyone has to change their HPA to `autoscalingv2beta2`. I mean the change from `metric.Object.MetricName` to `metric.Object.Metric.Name`.

arjunrn:
Closing in favor of #27

PR #27, opened 2019-01-10
Added weighting of RPS metrics based on backend weights

arjunrn:
# Weighting RPS metrics based on weight annotations

## Description
The backend weights for skipper ingress are stored as annotations in the ingress object. These are used to weight the returned metrics. Also the `request-per-second` metric name also contains the backend name.

## Types of Changes
- Weighting of RPS metrics
- Bug fix for when the MaxCollector gets only zero values

## Review
- [x] Tests
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,16 @@ spec:
       targetValue: 10 # this will be treated as targetAverageValue
 ```
 
+### Metric weighting based on backend
+
+Skipper supports sending traffic to different backend based on annotations present on the
+`Ingress` object. When the metric name is specified without a backend as `requests-per-second`
+then the number of replicas will be calculated based on the full traffic served by that ingress.
+If however only the traffic being routed to a specific backend should be used then the
+backend name can be specified as a metric name like `requests-per-second|backend1` which would
+return the requests-per-second being sent to the `backend1`.
+
+
 **Note:** As of Kubernetes v1.10 the HPA does not support `targetAverageValue` for
 metrics of type `Object`. In case of requests per second it does not make sense
 to scale on a summed value because you can not make the total requests per
--- a/README.md
+++ b/README.md
@@ -246,8 +246,9 @@ Skipper supports sending traffic to different backend based on annotations prese
 `Ingress` object. When the metric name is specified without a backend as `requests-per-second`
 then the number of replicas will be calculated based on the full traffic served by that ingress.
 If however only the traffic being routed to a specific backend should be used then the
-backend name can be specified as a metric name like `requests-per-second|backend1` which would
-return the requests-per-second being sent to the `backend1`.
+backend name can be specified as a metric name like `requests-per-second/backend1` which would
+return the requests-per-second being sent to the `backend1`. The ingress annotation where
+the backend weights can be obtained can be specified through the flag `--skipper-backends-annotation`.
 
 
 **Note:** As of Kubernetes v1.10 the HPA does not support `targetAverageValue` for
--- a/pkg/collector/skipper_collector.go
+++ b/pkg/collector/skipper_collector.go
@@ -1,20 +1,24 @@
 package collector
 
 import (
+	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"math"
 	"strings"
 	"time"
 
 	autoscalingv2beta1 "k8s.io/api/autoscaling/v2beta1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -240,6 +240,16 @@ spec:
       targetValue: 10 # this will be treated as targetAverageValue
 ```
 
+### Metric weighting based on backend
+
+Skipper supports sending traffic to different backend based on annotations present on the
```
szuecs: Maybe link skipper, if it's not done somewhere else, because it should not be implicit.

szuecs:
:+1:

README.md
arjunrn: It's already linked in an earlier section of the README.

README.md
```diff
@@ -240,6 +240,16 @@ spec:
...
+`Ingress` object. When the metric name is specified without a backend as `requests-per-second`
+then the number of replicas will be calculated based on the full traffic served by that ingress.
+If however only the traffic being routed to a specific backend should be used then the
+backend name can be specified as a metric name like `requests-per-second|backend1` which would
```
mikkeloscar: Is this correct? `requests-per-second|backend1` in the code it's splitting at `,` afaict. I dind't know you could even use characters like this I thought it was limited to alphanumeric plus `-` or something like that.

pkg/collector/skipper_collector.go
```diff
@@ -1,20 +1,24 @@
...
 	"k8s.io/metrics/pkg/apis/custom_metrics"
 )
 
 const (
-	rpsQuery      = `scalar(sum(rate(skipper_serve_host_duration_seconds_count{host="%s"}[1m])))`
-	rpsMetricName = "requests-per-second"
+	rpsQuery                  = `scalar(sum(rate(skipper_serve_host_duration_seconds_count{host="%s"}[1m])))`
+	rpsMetricName             = "requests-per-second"
+	stacksetWeightsAnnotation = "zalando.org/stack-traffic-weights"
+	backendWeightsAnnotation  = "zalando.org/backend-weights"
```
mikkeloscar: I'm not sure it makes sense to use a max over these two values as the weight because `zalando.org/stack-traffic-weights` is only the desired value while `zalando.org/backend-weights` is the actual current value and therefore more precisely represents the weight that the particular stack will be getting. Using only `zalando.org/backend-weights` would also limit this project to only depend on skipper and not have any relation to stackset-controller (only the other way around which is fine).

WDYT?

README.md
arjunrn: Yep, actually I left it like this to ask you which separator we should use. I ran my tests using the comma separator and it worked. It's just the name of the metric and I don't think it follows any rules like k8s resource rule names.

pkg/collector/skipper_collector.go
arjunrn: It does rely on skipper doesn't it? Because it's unlikely that anything else will expose those metrics with that name or is it a standard. Also I take the max over the two because say while scaling up the stackset-controller annotation is changed, but the skipper pods are still using the backend annotations. Scaling down the old stackset at that point may result in requests being dropped. That said it may make sense to pass the annotations as flags instead of hardcoding it as constants.

szuecs: Pass a flag makes most sense and set a default to `zalando.org/backend-weights`. Like this it works out of the box with skipper, but also traefik users can use it with their custom annotation.

README.md
szuecs: I like `,` separators more than `|`.

mikkeloscar: What about `/` if it's an option?

pkg/collector/skipper_collector.go
arjunrn: If we use only `backend-weights` won't the scale-up only happen once the traffic has been switched? I think it may cause issues because existing pods won't be able to handle the new load.

README.md
arjunrn: Slashed are also possible I guess.

pkg/collector/skipper_collector.go
mikkeloscar: I just don't think we can rely purely on `max( stack-traffic-weights, backend-weights)` for pre-scaling because it will anyway take some time for the HPA to react, and the stack might get traffic before the HPA has scaled. This is currently handled by pre-scaling in stackset-controller, which i think is fine.

That being said `max( stack-traffic-weights, backend-weights)` should not hurt as it would in worst case scale a bit too high. So I'm fine with `max( stack-traffic-weights, backend-weights)` based on our discussion here! :+1:

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #29, opened 2019-01-18
Added missing logic and test case when weight of backend is 0

arjunrn:
The case when the backend weight was zero was not handled and it was falling back to the logic when the weight is not set. This PR addresses that issue and adds a corresponding test.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #31, opened 2019-01-23
Fix case when backend is not set

arjunrn:
When the backend is not specified in the metric name the weight is defaults to zero. It should instead be 1 otherwise the `request-per-second` metric is useless.


arjunrn:
:+1:

mikkeloscar:
:+1:

PR #33, opened 2019-01-31
Handle condition when backend weights only sometimes present

arjunrn:
# Handle condition when backend weights only sometimes present

## Description
The case when the backend weight is present only in one of the ingress annotations was not handled which caused the backend weight to be calculated as 1. 

## Types of Changes

## Review
- [x] Tests

### DIFF
--- a/pkg/collector/skipper_collector.go
+++ b/pkg/collector/skipper_collector.go
@@ -81,29 +81,33 @@ func NewSkipperCollector(client kubernetes.Interface, plugin CollectorPlugin, hp
 	}, nil
 }
 
-func getAnnotationWeight(backendWeights string, backend string) float64 {
+func getAnnotationWeight(backendWeights string, backend string) (float64, bool) {
 	var weightsMap map[string]int
 	err := json.Unmarshal([]byte(backendWeights), &weightsMap)
 	if err != nil {
-		return 1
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

pkg/collector/skipper_collector.go
```diff
@@ -81,29 +81,33 @@ func NewSkipperCollector(client kubernetes.Interface, plugin CollectorPlugin, hp
...
 	var maxWeight float64 = -1
+	weightSet := false
 	for _, anno := range backendAnnotations {
 		if weightsMap, ok := ingressAnnotations[anno]; ok {
-			weight := getAnnotationWeight(weightsMap, backend)
-			if weight > maxWeight {
-				maxWeight = weight
+			weight, isPresent := getAnnotationWeight(weightsMap, backend)
+			if isPresent {
+				weightSet = true
```
szuecs: I think you don’t need to have isPresent to set weightSet :)

arjunrn: Actually I do. The isPresent is set in a loop and I need to know if the weight was set at least once because the code will have to return different results if the weight is set at least once. Or not at all.

arjunrn:
:+1:

PR #43, opened 2019-04-04
Fix nil dereference panics for the Annotation config parser

arjunrn:
The annotation config parser was not initialized correctly.

### DIFF
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -174,7 +174,7 @@ func ParseHPAMetrics(hpa *autoscalingv2.HorizontalPodAutoscaler) ([]*MetricConfi
 
 	// TODO: validate that the specified metric names are defined
 	// in the HPA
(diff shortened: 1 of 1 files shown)


pkg/collector/collector.go
```diff
@@ -174,7 +174,7 @@ func ParseHPAMetrics(hpa *autoscalingv2.HorizontalPodAutoscaler) ([]*MetricConfi
 
 	// TODO: validate that the specified metric names are defined
 	// in the HPA
-	var parser annotations.AnnotationConfigMap
+	parser := annotations.NewAnnotationConfigMap()
```
mikkeloscar: Maybe just `parser := make(annotations.AnnotationConfigMap)` instead of a dedicated function only calling make? :)

arjunrn:
:+1:

mikkeloscar:
:+1:

muaazsaleem:
👍

PR #44, opened 2019-04-05
Fixed metric labels so that metric are tagged correct

arjunrn:
Fixed metric labels so that metric are tagged correct. Also added nil check.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #47, opened 2019-04-26
Return a value when atleast one of the metrics returns a value

arjunrn:
The `MaxCollector` fails when even one of the metrics collector returns an error. But when other metrics are present it should just continue and return the max of the rest.


mikkeloscar:
Should we add a unit test? :)

muaazsaleem:
👍

muaazsaleem:
👍

arjunrn:
:+1:

PR #48, opened 2019-04-26
Remove unused Configuration key from MetricConfig

mikkeloscar:
This fixes an issue of setting up a ZMON collector where the incorrect
key `Configuration` was used, which was not initialized in the metrics
config parser. The `Config` key is the right one to use.


mikkeloscar:
:+1:

arjunrn:
:+1:

PR #50, opened 2019-04-27
Max collector should ignore only no result errors

arjunrn:
PR #47 has faulty logic. All errors including ones where the _metrics-adpater_ cannot connect to _prometheus_ are ignored. Only invalid results should be so if there is a connection error then there is no result computed.

### DIFF
--- a/pkg/collector/prometheus_collector.go
+++ b/pkg/collector/prometheus_collector.go
@@ -16,6 +16,14 @@ import (
 	"k8s.io/metrics/pkg/apis/custom_metrics"
 )
 
+type NoResultError struct {
+	query string
+}
+
+func (r NoResultError) Error() string {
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

pkg/collector/prometheus_collector.go
```diff
@@ -16,6 +16,14 @@ import (
 	"k8s.io/metrics/pkg/apis/custom_metrics"
 )
 
+type noResultError struct {
```
mikkeloscar: nit: Probably makes sense to expose the type as it's returned by the public `GetMetrics` method.

mikkeloscar:
:+1:

arjunrn:
👍

PR #53, opened 2019-04-30
Allow Prometheus metrics for External target

### LINKED ISSUE #45: Prometheus collector should create External target.type
## Expected Behavior
I think it makes more sense for the Prometheus collector to create an External metric instead of an Object. The syntax fits better. Also, although it can provide metrics about an Object, I doubt it will be used much in that way.

## Actual Behavior
It currently provides the metrics of type Object. So a dummy target has to be added and for Kubernetes <1.12 targetAverageValue requires a workaround.

mikkeloscar:
Allows specifying Prometheus based metrics as `External` target.

I'm not 100% sure on the format for the annotation. Feedback welcome.

* [x] Still need to test that it actually works this way :)

Fix #45

/cc @timstoop

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,9 +152,51 @@ the trade-offs between the two approaches.
 
 | Metric | Description | Type | Kind |
 | ------------ | -------------- | ------- | -- |
+| `prometheus-query` | Generic metric which requires a user defined query. | External | |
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

szuecs:
👍

PR #59, opened 2019-05-17
fix travis should fail fast if golangci-list can not be fetched

szuecs:
Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,12 @@ dist: xenial
 
 go:
   - "1.11.x"
-
-env:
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

mikkeloscar:
The error form the other build was:
```
curl: (7) Failed to connect to install.goreleaser.com port 443: Connection timed out
```

When downloading the binary, so the PATH is not the issue

szuecs:
I give up, stupid dependencies

PR #68, opened 2019-07-27
Add support for averageValue for request-per-second Skipper metric

mikkeloscar:
This adds support for `averageValue` for the `request-per-second` metric based on Ingress Objects. This is only supported from Kubernetes `>=v1.14` (https://github.com/kubernetes/kubernetes/pull/72872).

When defining the HPA with `autoscaling/v2beta1` you still need to define `targetValue` even though it won't be used when `averageValue` is set. Once we default to `autoscaling/v2beta2` this akward API will be gone.

This PR also enhances the README with Kubernetes compatibility expectations and a reference to supported Kubernetes versions for each collector type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,12 @@
 # kube-metrics-adapter
 [![Build Status](https://travis-ci.org/zalando-incubator/kube-metrics-adapter.svg?branch=master)](https://travis-ci.org/zalando-incubator/kube-metrics-adapter)
 [![Coverage Status](https://coveralls.io/repos/github/zalando-incubator/kube-metrics-adapter/badge.svg?branch=master)](https://coveralls.io/github/zalando-incubator/kube-metrics-adapter?branch=master)
+
 Kube Metrics Adapter is a general purpose metrics adapter for Kubernetes that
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -284,7 +299,10 @@ spec:
         apiVersion: extensions/v1beta1
         kind: Ingress
         name: myapp
-      targetValue: 10 # this will be treated as targetAverageValue
+      averageValue: 10 # Only works with Kubernetes >=1.14
```
arjunrn: It's called `targetAverageValue`

mikkeloscar: Not here https://github.com/kubernetes/kubernetes/blob/a87e9a978f65a8303aa9467537aa59c18122cbf9/staging/src/k8s.io/api/autoscaling/v2beta1/types.go#L135 :)

It would make too much sense if the names matched across metric types 😂

arjunrn: I don't understand why they have versioned betas if they keep changing all the versions. Oh well... 😄

mikkeloscar:
:+1:

mikkeloscar:
:+1:

szuecs:
👍

PR #70, opened 2019-07-30
RPS: Correctly handle backends not present in the annotations

aermakov-zalando:
Backends with zero weights are completely missing from the traffic weight annotations. This currently triggers the fallback logic in the adapter and they're treated as backends that receive 100% of the traffic, causing massive scale ups. Fix by explicitly checking if we've found any of the annotations instead.


aermakov-zalando:
👍

arjunrn:
👍

PR #71, opened 2019-07-30
Skipper: fix the no annotation test so it makes more sense

aermakov-zalando:
We want to check that there are no annotations on the ingress, not in the configuration.


aermakov-zalando:
👍

arjunrn:
👍

PR #78, opened 2019-09-27
When traffic switching is used, require a backend for the RPS metric

aermakov-zalando:
In #70 we changed the logic slightly to essentially require a backend name for stacks with traffic switching, but didn't return an error. This doesn't allow the users notice that their HPAs are misconfigured, they just don't work. Fix by correctly returning an error.


arjunrn:
👍

aermakov-zalando:
👍

PR #79, opened 2019-09-30
Fail on dirty and/or non-exact versions on master

aermakov-zalando:
* Fail on dirty repositories
 * Fail if we don't have a version tag on `master`

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -20,7 +20,7 @@ pipeline:
     cmd: |
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         IMAGE=registry-write.opensource.zalan.do/teapot/kube-metrics-adapter
-        VERSION=$(git describe --tags --exact)
+        VERSION=$(git describe --tags --always)
       else
         IMAGE=registry-write.opensource.zalan.do/teapot/kube-metrics-adapter-test
         VERSION=$CDP_BUILD_VERSION
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -7,6 +7,8 @@ pipeline:
     - /go/pkg/mod       # pkg cache for Go modules
     - ~/.cache/go-build # Go build cache
   type: script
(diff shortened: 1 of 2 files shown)


delivery.yaml
```diff
@@ -7,6 +7,8 @@ pipeline:
     - /go/pkg/mod       # pkg cache for Go modules
     - ~/.cache/go-build # Go build cache
   type: script
+  env:
+    GOFLAGS: "-mod=readonly"
```
szuecs: you should just move forward to Go 1.13 and commit the change of go.mod

aermakov-zalando:
👍

aermakov-zalando:
👍

aermakov-zalando:
👍

delivery.yaml
aermakov-zalando: I did that, but I'd still like the build to just fail instead of silently doing things in the background. No idea why this isn't the default.

delivery.yaml
```diff
@@ -18,9 +20,11 @@ pipeline:
     cmd: |
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         IMAGE=registry-write.opensource.zalan.do/teapot/kube-metrics-adapter
-        VERSION=$(git describe --tags --always --dirty)
+        VERSION=$(git describe --tags --exact)
```
mikkeloscar: Does this mean we can't build master images for non-tagged versions? It's IMO convenient to be able to try an image out before tagging the version.

aermakov-zalando: Shouldn't the PR versions be enough for this?

aermakov-zalando: I don't mind reverting, though, but I'd still like to keep the `--dirty` part dropped.

aermakov-zalando: Fixed.

aermakov-zalando:
👍

szuecs:
:+1:

PR #83, opened 2019-10-21
Skipper: simplify metrics collection

aermakov-zalando:
* Drop MaxWeightedCollector (we don't want max anyway, we want sum)
 * Use Prometheus to add up all matching metrics and scale them; this has a nice side effect of ensuring that unused hostnames don't cause an error when collecting the metrics
 * Update the tests a bit


aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #85, opened 2019-10-22
Support for JSONPath expressions that return arrays of values

tomaspinho:
This is the initial implementation of support for JSONPath expressions that return arrays of values instead of a single value.

# One-line summary
This is the initial implementation of support for JSONPath expressions that return arrays of values instead of a single value.

## Description

This extends the collector to define a few handy reducer functions that take in the slice of float64 and return a single value. It also allows the user to define which reducer function to use via the `metric-config.<metricType>.<metricName>.json-path/reducer-func` annotation, which can have the values of 'avg', 'min', 'max' and 'sum'.

For instance, the Ruby puma webserver exposes metrics of the form of `$.worker_status[*].last_status.pool_capacity` that have to be consumed as an array of values to be properly targetted.

## Types of Changes

## Tasks
  - [x] Include documentation in the README for new annotation

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
NA

### DIFF
--- a/pkg/collector/json_path_collector.go
+++ b/pkg/collector/json_path_collector.go
@@ -146,16 +146,18 @@ func getPodMetrics(pod *corev1.Pod, scheme, path string, port int) ([]byte, erro
 
 // intsToFloat64s will convert a slice of int to a slice of float64
 func intsToFloat64s(in []int) (out []float64) {
-	for i, v := range in {
-		out[i] = float64(v)
+	out = []float64{}
+	for _, v := range in {
+		out = append(out, float64(v))
 	}
 	return
 }
 
 // float32sToFloat64s will convert a slice of float32 to a slice of float64
 func float32sToFloat64s(in []float32) (out []float64) {
-	for i, v := range in {
-		out[i] = float64(v)
+	out = []float64{}
+	for _, v := range in {
+		out = append(out, float64(v))
 	}
 	return
 }
--- a/pkg/collector/json_path_collector.go
+++ b/pkg/collector/json_path_collector.go
@@ -162,7 +162,7 @@ func float32sToFloat64s(in []float32) (out []float64) {
 	return
 }
 
-// reduce will reduce a slice of numbers given a aggregator function's name. If it's empty or not recognized, avg is used.
+// reduce will reduce a slice of numbers given a aggregator function's name. If it's empty or not recognized, an error is returned.
 func reduce(values []float64, aggregator string) (float64, error) {
 	switch aggregator {
 	case "avg":
--- a/pkg/collector/json_path_collector.go
+++ b/pkg/collector/json_path_collector.go
@@ -18,11 +18,11 @@ import (
 // querying the pods metrics endpoint and lookup the metric value as defined by
 // the json path query.
 type JSONPathMetricsGetter struct {
-	jsonPath    *jsonpath.Compiled
-	scheme      string
-	path        string
-	port        int
-	reducerFunc string
+	jsonPath   *jsonpath.Compiled
+	scheme     string
+	path       string
+	port       int
+	aggregator string
 }
 
 // NewJSONPathMetricsGetter initializes a new JSONPathMetricsGetter.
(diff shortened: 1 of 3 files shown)


pkg/collector/json_path_collector.go
```diff
@@ -131,3 +143,80 @@ func getPodMetrics(pod *corev1.Pod, scheme, path string, port int) ([]byte, erro
...
+	case "avg":
+		return avg(values)
+	case "min":
+		return min(values)
+	case "max":
+		return max(values)
+	case "sum":
+		return sum(values)
+	default:
+		return avg(values)
```
arjunrn: I would rather this function return an error than use average. Currently misconfigurations would be silently handled.

arjunrn:
@tomaspinho thank you for the PR. Overall looks good other than one small change. Also do you mind adding tests. I know there are none at the moment but it should not be too difficult to add some. If not that's also ok for now.

pkg/collector/json_path_collector.go
```diff
@@ -17,10 +18,11 @@ import (
...
 type JSONPathMetricsGetter struct {
-	jsonPath *jsonpath.Compiled
-	scheme   string
-	path     string
-	port     int
+	jsonPath    *jsonpath.Compiled
+	scheme      string
+	path        string
+	port        int
+	reducerFunc string
```
mikkeloscar: In other collectors we call `reducerFunc` for `aggregators` (terminology used in KairosDB). Could we use the same terminology here to be consistent? :)

tomaspinho: Just pushed a change to rename `reducerFunc` to `aggregator` for consistency purposes!

pkg/collector/json_path_collector.go
tomaspinho: Pushed a change to implement this!

tomaspinho:
Writing some tests to test non-mockable sections of my code, not too comfortable with testing original code.

arjunrn:
> Writing some tests to test non-mockable sections of my code, not too comfortable with testing original code.

@tomaspinho everything after [this](https://github.com/zalando-incubator/kube-metrics-adapter/blob/c006ef010e67ff03e848b9db687ec8ea43bdfd8d/pkg/collector/json_path_collector.go#L74) doesn't have dependencies. You could move the code after this into a function called `parseData` and then write tests for only the functionality you added.

tomaspinho:
@arjunrn ended up testing all functions in file, excluding `GetMetric`. Coverage increased by 2.7% globally. What do you think?

arjunrn:
@tomaspinho thank you for the contribution. Looks good to me.

arjunrn:
:+1:

pkg/collector/json_path_collector.go
```diff
@@ -131,3 +143,82 @@ func getPodMetrics(pod *corev1.Pod, scheme, path string, port int) ([]byte, erro
...
+func float32sToFloat64s(in []float32) (out []float64) {
+	out = []float64{}
+	for _, v := range in {
+		out = append(out, float64(v))
+	}
+	return
+}
+
+// reduce will reduce a slice of numbers given a reducer function's name. If it's empty or not recognized, avg is used.
+func reduce(values []float64, aggregator string) (float64, error) {
```
mikkeloscar: nit: the comment here is now a bit outdated after the name change to `aggregator`

tomaspinho: fixed! :D

mikkeloscar: The default also changed from `avg` to `error` right? :)

mikkeloscar: Sorry for nitpicking. I think it's the only thing, then it's gtg from my side.

tomaspinho: No worries, it's my fault :)

tomaspinho: Fixed in last commit

mikkeloscar:
:+1:

szuecs:
:+1:

PR #87, opened 2019-10-25
Fix, use collector name to get plugin, not metric name

pinkavaj:
## Description
For external metrics was used metric name to get collector, but this does not correspond to documentation. In example configuration of prometheus is used collector name not metric name.

Other possibility is, the documentation is wrong (or i missed something :)

This MR demonstrates the problem, but brings backward incopatibile change! Serious thinking should be done before considering merging this. Mey-be another approach would be more appropriate (keep somehow bacward compatibility?).

## Types of Changes

- Bug fix (__breaking__ change which fixes an issue)

## Tasks
Verify original author intentions.

## Review
- [ ] Tests
- [ ] CHANGELOG

### DIFF
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -134,7 +134,7 @@ func (c *CollectorFactory) NewCollector(hpa *autoscalingv2.HorizontalPodAutoscal
 			return c.objectPlugins.Any.Any.NewCollector(hpa, config, interval)
 		}
 	case autoscalingv2.ExternalMetricSourceType:
-		if plugin, ok := c.externalPlugins[config.Metric.Name]; ok {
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Hi, sorry for not looking into this sooner.

I don't see the need for this change as we don't really document how this is implemented. From my PoV it's an implementation detail which I don't see reflected in the user documentation. Right now with this change I believe we would break the AWS SQS metric because for that we don't have a "collectorName" as this extra information is not needed. So for this reason I wouldn't merge it as is.

I would be happy to make the documentation more clear, but it's just not clear to me where you get the assumption that "CollectorName" is documented for this?

mikkeloscar:
I will close this for now, but feel free to re-open to further discuss how we can make it less confusing in the docs.

Issue #94, opened 2019-11-26
Doesn't play well with another collector


tomaspinho:
## Expected Behavior
`kube-metrics-adapter` should behave well in an environment with a different external or custom metrics provider.

## Actual Behavior
`kube-metrics-adapter` will attempt modifying HPAs that have external or custom metrics when it is not configured as a metrics provider for that metric type.

Logs:
```
time="2019-11-26T10:00:23Z" level=info msg="Event(v1.ObjectReference{Kind:\"HorizontalPodAutoscaler\", Namespace:\"development\", Name:\"task-worker\", UID:\"8e7d8d12-0c84-11ea-8744-0a26919e3b5a\", APIVersion:\"autoscaling/v2beta2\", ResourceVersion:\"92543736\", FieldPath:\"\"}): type: 'Warning' reason: 'CreateNewMetricsCollector' Failed to create new metrics collector: no plugin found for {External {aws.sqs.sqs_messages_visible &LabelSelector{MatchLabels:map[string]string{aws_account_name: development,queue_name: tasks,region: eu-west-1,},MatchExpressions:[]LabelSelectorRequirement{},}}}"
```

Event on respective HPA:
```
  Type     Reason                     Age                   From                  Message
  ----     ------                     ----                  ----                  -------
  Warning  CreateNewMetricsCollector  94s (x1947 over 16h)  kube-metrics-adapter  Failed to create new metrics collector: no plugin found for {External {aws.sqs.sqs_messages_visible &LabelSelector{MatchLabels:map[string]string{aws_account_name: development,queue_name: tasks,region: eu-west-1,},MatchExpressions:[]LabelSelectorRequirement{},}}}
```

## Steps to Reproduce the Problem

  1. Install this project only as a Custom Metrics Provider
  2. Install `datadog-cluster-agent` as an External Metrics Provider, for instance. 
  3. Configure a HPA targetting External Metrics

## Specifications

  - Version: v0.0.5
  - Platform: Kubernetes
  - Subsystem: HPA provider

tomaspinho:
Is there a surefire way of introducing logic to the HPA provider to only target HPAs for which the collector is configured for? Or that have any annotations beginning with `metric-config`?

mikkeloscar:
What is the actual impact you observe? Does it just print errors because it can't enable a collector for the HPA, or does it cause actual issues?

If it's just an error we can maybe maybe it less verbose?

tomaspinho:
[...]

PR #95, opened 2019-11-27
Permit disregarding incompatible HPAs

tomaspinho:
# One-line summary
Permit disregarding incompatible HPAs.

> Issue : #94

## Description
This adds a `--disregard-incompatible-hpas` that makes the HPA provider stop erroring out when a collector cannot be created for a metric in a HPA. Useful when kube-metrics-adapter runs alongside another metrics provider. Fixes issue #94.

## Types of Changes

- Configuration change

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
NA. The flag should assume the sensible Go zero value default of `false`.

### DIFF
--- a/pkg/provider/hpa.go
+++ b/pkg/provider/hpa.go
@@ -168,7 +168,8 @@ func (p *HPAProvider) updateHPAs() error {
 				c, err := p.collectorFactory.NewCollector(&hpa, config, interval)
 				if err != nil {
 
-					if errors.Is(err, &collector.PluginNotFoundError{}) && !p.disregardIncompatibleHPAs {
+					// Only log when it's not a PluginNotFoundError AND flag disregardIncompatibleHPAs is true
+					if !(errors.Is(err, &collector.PluginNotFoundError{}) && p.disregardIncompatibleHPAs) {
 						p.recorder.Eventf(&hpa, apiv1.EventTypeWarning, "CreateNewMetricsCollector", "Failed to create new metrics collector: %v", err)
 					}
 
--- a/pkg/provider/hpa.go
+++ b/pkg/provider/hpa.go
@@ -49,16 +49,17 @@ var (
 // HPAProvider is a base provider for initializing metric collectors based on
 // HPA resources.
 type HPAProvider struct {
-	client             kubernetes.Interface
-	interval           time.Duration
-	collectorScheduler *CollectorScheduler
-	collectorInterval  time.Duration
-	metricSink         chan metricCollection
-	hpaCache           map[resourceReference]autoscalingv2.HorizontalPodAutoscaler
-	metricStore        *MetricStore
-	collectorFactory   *collector.CollectorFactory
-	recorder           kube_record.EventRecorder
-	logger             *log.Entry
+	client                    kubernetes.Interface
+	interval                  time.Duration
+	collectorScheduler        *CollectorScheduler
+	collectorInterval         time.Duration
+	metricSink                chan metricCollection
+	hpaCache                  map[resourceReference]autoscalingv2.HorizontalPodAutoscaler
+	metricStore               *MetricStore
+	collectorFactory          *collector.CollectorFactory
+	recorder                  kube_record.EventRecorder
+	logger                    *log.Entry
+	disregardIncompatibleHPAs bool
 }
 
 // metricCollection is a container for sending collected metrics across a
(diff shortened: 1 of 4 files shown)


tomaspinho:
@arjunrn sorry for the direct mention, but any news on this? Anything else needed from my side? Thanks :smile:

pkg/provider/hpa.go
```diff
@@ -163,7 +165,7 @@ func (p *HPAProvider) updateHPAs() error {
 				}
 
 				collector, err := p.collectorFactory.NewCollector(&hpa, config, interval)
-				if err != nil {
+				if err != nil && !p.disregardIncompatibleHPAs {
```
mikkeloscar: If you get an error here and have `disregardIncompatibleHPAs==true` then we would try to add the `nil` collector below.

Better would be to determine from the error if it was an error setting up the collector which should not happen or if the error was simply because no collector is supported for the HPA.

tomaspinho: I have refactored this to address your concerns. What do you think if the latest commit?

pkg/server/start.go
```diff
@@ -104,7 +104,8 @@ func NewCommandStartAdapterServer(stopCh <-chan struct{}) *cobra.Command {
 		"whether to enable AWS external metrics")
 	flags.StringSliceVar(&o.AWSRegions, "aws-region", o.AWSRegions, "the AWS regions which should be monitored. eg: eu-central, eu-west-1")
 	flags.StringVar(&o.MetricsAddress, "metrics-address", o.MetricsAddress, "The address where to serve prometheus metrics")
-
+	flags.BoolVar(&o.DisregardIncompatibleHPAs, "disregard-incompatible-hpas", o.DisregardIncompatibleHPAs, ""+
+		"whether to disregard failing to create collectors for incompatible HPAs")
```
arjunrn: The _whether to_ is redundant here

arjunrn:
@mikkeloscar @tomaspinho My only concern is that if a user makes a typo for say a custom metric and the collector is not found then the adapter will basically swallow the error when the `--disregard-compatible-hpa` flag is set. Correct me if I am wrong.

pkg/provider/hpa.go
```diff
@@ -162,15 +165,19 @@ func (p *HPAProvider) updateHPAs() error {
 					interval = p.collectorInterval
 				}
 
-				collector, err := p.collectorFactory.NewCollector(&hpa, config, interval)
+				c, err := p.collectorFactory.NewCollector(&hpa, config, interval)
 				if err != nil {
-					p.recorder.Eventf(&hpa, apiv1.EventTypeWarning, "CreateNewMetricsCollector", "Failed to create new metrics collector: %v", err)
+
+					if errors.Is(err, &collector.PluginNotFoundError{}) && !p.disregardIncompatibleHPAs {
```
mikkeloscar: We should be able to log errors which are NOT `PluginNotFoundError`. Currently those are not reported.

mikkeloscar:
@arjunrn IMO it's fine, we can't really decide if it's a typo or the name of a metric handled by another metrics provider.

pkg/provider/hpa.go
tomaspinho: Indeed, the latest commit addresses this. Even did a truth table on paper to make sure the expression was correct.

mikkeloscar: Doesn't seem right. Now we would only log if `disregardIncompatibleHPAs` is enabled.

We should log any error when `disregardIncompatibleHPAs=false` and log all errors which is not `PluginNotFoundError` when `disregardIncompatibleHPAs=true`.

tomaspinho: Ok, let me reproduce my truth table here:

Log = !(errors.Is(err, &collector.PluginNotFoundError{}) && p.disregardIncompatibleHPAs)

| Log | is PluginNotFoundError | disregardIncompatibleHPAs = true |
|-|-|-|
|F|T|T|
|T|T|F|
|T|F|T|
|T|F|F|

mikkeloscar: OK, I missed the wrapping `!()`. It seems a bit hard to read this way, but as it's correct I can live with it :)

tomaspinho: nice! 😄

tomaspinho:
I'll try to improve test coverage as it seems like the only thing blocking this.

pkg/provider/hpa_test.go
```diff
@@ -90,3 +90,51 @@ func TestUpdateHPAs(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	fakeClient := fake.NewSimpleClientset()
+
+	var err error
+	hpa, err = fakeClient.AutoscalingV2beta1().HorizontalPodAutoscalers("default").Create(hpa)
```
golangcibot: ineffectual assignment to `hpa` (from `ineffassign`)

tomaspinho:
Test coverage has been improved 😄

mikkeloscar:
:+1:

arjunrn:
:+1:

PR #102, opened 2020-01-30
Reuse the HTTP client for scraping pods

arjunrn:
Currently a new `http.Client` is created whenever a pod metrics endpoint has to be scraped. This leaves idle connections lying around which piles up and causes a memory leak. By reusing the client and setting idle connection timeouts we can prevent this.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	_ "net/http/pprof"
 	"os"
 	"runtime"
 
--- a/pkg/collector/json_path_collector.go
+++ b/pkg/collector/json_path_collector.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math"
+	"net"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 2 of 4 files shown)


mikkeloscar:
We still create an HTTP client for reach collection loop (instead of one * pods per collection loop) so I think the idle timeout is the biggest improvement.

Let's go with this for now, but maybe it makes sense to think about reusing the HTTP clients across collection loops

mikkeloscar:
:+1:

arjunrn:
@mikkeloscar The `JSONPathMetricsGetter` is created along with the `PodCollector`. So it's created only once when the HPA is discovered or changed, right?

mikkeloscar:
@arjunrn You're right :man_facepalming:

arjunrn:
:+1:

main.go
```diff
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	_ "net/http/pprof"
```
aermakov-zalando: Do we need this, or was this temporary for debugging?

arjunrn: I left it in deliberately. I don't think it should have any impact because this is not a performance critical application.

PR #107, opened 2020-02-21
fix(influxdb): fix swap of org/token

affo:
## One line summary

Fix an error in passing parameters to InfluxDBCollector and convert to `org` instead of `orgID` (that's actually how InfluxDB treats that param, it could be the org name or ID).

## Description

Fixing that error would make it work 😅 

## Types of Changes

- Refactor/improvements
- Documentation / non-code

### DIFF
--- a/pkg/collector/influxdb_collector.go
+++ b/pkg/collector/influxdb_collector.go
@@ -18,7 +18,7 @@ const (
 	InfluxDBMetricName        = "flux-query"
 	influxDBAddressKey        = "address"
 	influxDBTokenKey          = "token"
-	influxDBOrgIDKey          = "org-id"
+	influxDBOrgKey            = "org"
 	influxDBQueryNameLabelKey = "query-name"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/collector/influxdb_collector.go
```diff
@@ -18,34 +18,34 @@ const (
...
 		address:    address,
 		token:      token,
-		orgID:      orgID,
+		org:        org,
 	}, nil
 }
 
 func (p *InfluxDBCollectorPlugin) NewCollector(hpa *v2beta2.HorizontalPodAutoscaler, config *MetricConfig, interval time.Duration) (Collector, error) {
-	return NewInfluxDBCollector(p.address, p.orgID, p.token, config, interval)
+	return NewInfluxDBCollector(p.address, p.token, p.org, config, interval)
```
affo: this is the actual bug fix. The rest is refactoring.

arjunrn:
👍

mikkeloscar:
:+1:

Issue #149, opened 2020-05-07
Support for multiple external http metrics.


lomkju:
Hi,

Can we somehow get metrics from multiple external HTTP endpoints.
Something like below
```
apiVersion: autoscaling/v2beta2
kind: HorizontalPodAutoscaler
metadata:
  name: echo-server
  annotations:
    # metric-config.<metricType>.<metricName>.<collectorName>/<configKey>
    metric-config.external.http-0.json/json-key: "$stats"
    metric-config.external.http-0.json/endpoint: "http://www.mocky.io/v2/5eb3fd280e0000670008180c"
    metric-config.external.http-1.json/json-key: "$stats"
    metric-config.external.http-1.json/endpoint: "http://www.mocky.io/v2/5eb3fd280e0000670008180c"
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: echo-server
  minReplicas: 1
  maxReplicas: 10
  metrics:
  - type: External
    external:
      metric:
        name: http-0
        selector:
          matchLabels:
            identifier: kafka_queue_length
      target:
        averageValue: 1
        type: AverageValue
  - type: External
    external:
      metric:
        name: http-1
        selector:
          matchLabels:
            identifier: kafka_ingest_rate
      target:
        averageValue: 2
        type: AverageValue

```

arjunrn:
@lomkju This should work out of the box. What's the error message you get?

lomkju:
The metric name needs to be http for it to work.
Below is the error for the same.
```
time="2020-05-08T21:05:38Z" level=info msg="Found 1 new/updated HPA(s)" provider=hpa
time="2020-05-08T21:05:38Z" level=info msg="Event(v1.ObjectReference{Kind:\"HorizontalPodAutoscaler\", Namespace:\"arjun-test\", Name:\"echo-server\", UID:\"e61d7a47-e6cf-4efe-bae9-ad2538d0cc3f\", APIVersion:\"autoscaling/v2beta2\", ResourceVersion:\"33712466\", FieldPath:\"\"}): type: 'Warning' reason: 'CreateNewMetricsCollector' Failed to create new metrics collector: no plugin found for {External {http-1 &LabelSelector{MatchLabels:map[string]string{identifier: kafka_ingest_rate,},MatchExpressions:[]LabelSelectorRequirement{},}}}"
[...]

PR #151, opened 2020-05-11
Allow multiple HTTP JSON external metrics

arjunrn:
# Allow override of HTTP JSON metrics with labels

> Issue : #149 

## Description
The latest version of the autoscaling API supports labels for the external metric sources. These could be leveraged to pass parameters for metrics collection.


arjunrn:
HPA controller does not like URLs and JSON paths in the label values:

```
  Warning  FailedGetExternalMetric       119s                  horizontal-pod-autoscaler  invalid label value: "$.stats": at key: "json-key": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

PR #152, opened 2020-05-14
Fetch pod metrics in parallel

doyshinda:
# One-line summary

Fetch metrics from pods in parallel

## Description
Fetching metrics from pods sequentially, with a large number of
pods, can result in poor performance when some of those pods have
been terminated by the HPA in a normal scale down event.

E.g., imagine the following sequence of events:
* kube-metrics-adapter retrieves a list of 30 pods to query for metrics
* HPA scales down to 20 pods
* kube-metrics-adapter attempts to retrieve metric from pod that is terminating/terminated, times out after 15 seconds
* 15 second timeout for other 9 terminating/terminated pods

2.5 minutes spent retrieving metrics

## Types of Changes
- Refactor/improvements

## Tasks
 - [x] Get all automated checks green
 - [ ] Get PR reviewed

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/pkg/collector/pod_collector.go
+++ b/pkg/collector/pod_collector.go
@@ -10,6 +10,7 @@ import (
 	autoscalingv2 "k8s.io/api/autoscaling/v2beta2"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/metrics/pkg/apis/custom_metrics"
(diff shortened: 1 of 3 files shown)


pkg/collector/pod_collector.go
```diff
@@ -85,32 +86,20 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
 		return nil, err
 	}
 
-	values := make([]CollectedMetric, 0, len(pods.Items))
-
-	// TODO: get metrics in parallel
+	ch := make(chan CollectedMetric)
+	err_ch := make(chan string)
```
mikkeloscar, line 90: I think this should be an `error` instead of `string`. As a small nitpick the variable name should be something like `errCh` to follow the go naming convention.

doyshinda: Will update the variable name!

I might be missing something, but what would the value be in returning the `error`? We need the `error` AND the pod information (Name + Namespace) to log an informative error message, but after that, the `error` isn't used anymore. If do return the `error` I guess I can just drop it in the `select`?.

Side note, this implies I'd have to log the error message inside to go routine (where the pod information is available) and I'm not sure if the logging library will correctly handle concurrent logging without messing up the format?

Anyways, happy to make whichever method is preferred here!

mikkeloscar: I simply mean it would be more semantically right to send an error type over the channel as opposed to a string. The error could still be constructed from `fmt.Errorf` and include all the relevant information.

You can also log directly from the go routine, this is not a problem. I understood that you didn't only use it for synchronizing the logging but also as a way to know what go routines where done or not.

doyshinda: Ah yes, I understand now! I thought you were referencing the `error` returned from `GetMetric()`. Will update.

mikkeloscar:
:+1:

arjunrn:
👍

PR #156, opened 2020-05-26
Add support for Ingress of API Group networking.k8s.io

mikkeloscar:
This adds support for ingresses under `networking.k8s.io/v1beta1` such that there is a migration path for ingresses currently under `extensions/v1beta1` for metrics relying on the skipper collector.

It also changes the read api for ingresses to use `networking.k8s.io/v1beta1` this is available since Kubernetes v1.14 which is already the minimum version supported for the skipper collector.


arjunrn:
:+1:

mikkeloscar:
:+1:

PR #162, opened 2020-06-26
Update client-go and custom-metrics-apiserver deps

mikkeloscar:
Just bumping the dependencies to the latest version.

PR #163, opened 2020-06-26
Gen openapi

### LINKED ISSUE #99: OpenAPI spec does not exist
Hi guys! More of an annoyance than anything else really, but would nice if we could avoid having the API server stop spewing logs about "OpenAPI spec does not exist" for the `v1beta1.external.metrics.k8s.io` and `v1beta1.custom.metrics.k8s.io` objects.

## Expected Behavior

API server logs not full of messages like:

```
1 controller.go:114] loading OpenAPI spec for "v1beta1.external.metrics.k8s.io" failed with: OpenAPI spec does not exist
1 controller.go:114] loading OpenAPI spec for "v1beta1.custom.metrics.k8s.io" failed with: OpenAPI spec does not exist
```

## Actual Behavior

API server logs full of messages like that.

I guess the best course of action would be to expose OpenAPI specs for those 2 objects - found this [commit elsewhere](https://github.com/kubernetes-sigs/metrics-server/pull/65/commits/06bcca5d06eae0a46fd1909a733e9f8be85b74d1) which seemed to detail a similar annoyance.

Anyway, thanks for kube-metrics-adapter!

mikkeloscar:
Depends on #162 

Adds support for serving OpenAPI spec.

Honestly I don't fully understand how this works, but it's simply modeled after: https://github.com/kubernetes-sigs/custom-metrics-apiserver/pull/65

Fix #99

Errors observed:

```
kubectl --namespace kube-system logs -f kube-metrics-adapter-7cf68b79f-fgv99
I0626 17:42:21.304236       1 serving.go:307] Generated self-signed cert (apiserver.local.config/certificates/apiserver.crt, apiserver.local.config/certificates/apiserver.key)
time="2020-06-26T17:42:21Z" level=info msg="Looking for HPAs" provider=hpa
F0626 17:42:21.801024       1 openapi.go:39] Failed to build open api spec for root: duplicate parameter labelSelector for route [{},{},{},{},{},{},{},{},{},{},{},{},{},{}], {"Name":"pretty","Description":"If 'true', then the output is pretty printed.","DataType":"string","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"labelSelector","Description":"A selector to restrict the list of returned objects by their labels. Defaults to everything.","DataType":"string","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"fieldSelector","Description":"A selector to restrict the list of returned objects by their fields. Defaults to everything.","DataType":"string","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"watch","Description":"Watch for changes to the described resources and return them as a stream of add, update, and remove notifications. Specify resourceVersion.","DataType":"boolean","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"allowWatchBookmarks","Description":"allowWatchBookmarks requests watch events with type \"BOOKMARK\". Servers that do not implement bookmarks may ignore this flag and bookmarks are sent at the server's discretion. Clients should not assume bookmarks are returned at any specific interval, nor may they assume the server will send any BOOKMARK event during a session. If this is not a watch, this field is ignored. If the feature gate WatchBookmarks is not enabled in apiserver, this field is ignored.","DataType":"boolean","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"resourceVersion","Description":"When specified with a watch call, shows changes that occur after that particular version of a resource. Defaults to changes from the beginning of history. When specified for list: - if unset, then the result is returned from remote storage based on quorum-read flag; - if it's 0, then we simply return what we currently have in cache, no guarantee; - if set to non zero, then the result is at least as fresh as given rv.","DataType":"string","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"timeoutSeconds","Description":"Timeout for the list/watch call. This limits the duration of the call, regardless of any activity or inactivity.","DataType":"integer","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"limit","Description":"limit is a maximum number of responses to return for a list call. If more items exist, the server will set the `continue` field on the list metadata to a value that can be used with the same initial query to retrieve the next set of results. Setting a limit may return fewer than the requested amount of items (up to zero items) in the event all requested objects are filtered out and clients should only use the presence of the continue field to determine whether more results are available. Servers may choose not to support the limit argument and will return all of the available results. If limit is specified and the continue field is empty, clients may assume that no more results are available. This field is not supported if watch is true.\n\nThe server guarantees that the objects returned when using continue will be identical to issuing a single list call without a limit - that is, no objects created, modified, or deleted after the first request is issued will be included in any subsequent continued requests. This is sometimes referred to as a consistent snapshot, and ensures that a client that is using limit to receive smaller chunks of a very large result can ensure they see all possible objects. If objects are updated during a chunked list the version of the object that was present at the time the first list result was calculated is returned.","DataType":"integer","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"continue","Description":"The continue option should be set when retrieving more results from the server. Since this value is server defined, clients may only use the continue value from a previous query result with identical query parameters (except for the value of continue) and the server may reject a continue value it does not recognize. If the specified continue value is no longer valid whether due to expiration (generally five to fifteen minutes) or a configuration change on the server, the server will respond with a 410 ResourceExpired error together with a continue token. If the client needs a consistent list, it must restart their list without the continue field. Otherwise, the client may send another list request with the token received with the 410 error, the server will respond with a list starting from the next key, but from the latest snapshot, which is inconsistent from the previous list results - objects that are created, modified, or deleted after the first list request will be included in the response, as long as their keys are after the \"next key\".\n\nThis field is not supported when watch is true. Clients may start a watch from the last resourceVersion value returned by the server and not miss any modifications.","DataType":"string","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
{"Name":"labelSelector","Description":"","DataType":"string","DataFormat":"","Kind":1,"Required":false,"AllowableValues":null,"AllowMultiple":false,"DefaultValue":"","CollectionFormat":"csv"}
```


mikkeloscar:
Doesn't work like this.

PR #173, opened 2020-07-01
Update MAINTAINERS

arjunrn:
Since I won't be able to dedicate time for this project and since I often get assigned issues/PR I though it would be right to remove myself.


mikkeloscar:
Thanks, can you also simply drop this file? https://github.com/zalando-incubator/kube-metrics-adapter/blob/master/.github/CODEOWNERS I don't want to be auto-assigned like you were :P, and I think this is the reason rather than the MAINTAINERS file.

mikkeloscar:
:+1:

gargravarr:
:+1:

PR #176, opened 2020-07-15
Correctly treat ingress weight values as float64

mikkeloscar:
The value of `zalado.org/backend-weights` annotation on an ingress is defined as `map[string]float64` in most places e.g skipper and stackset-controller. 
Before we parsed it as `map[string]int` which would not work if a weight was defined e.g. as `100.0` as it would result in the following error: `json: cannot unmarshal number 100.0 into Go value of type int`.

For most cases this was actually not a problem as we mostly use whole numbers and go will marshal a `float64` like `100.0` into the string representation `100` which can be unmarshaled into an `int`.
However for some ingresses the values are represented as `100.0` and this would fail.

* This changes both the type from `int` -> `float64` and properly handles any json parsing errors so such issues wouldn't be swallowed without a clear reason in the future.
* Changes a test case to use decimal numbers to validate the parsing of floats.

### DIFF
--- a/pkg/collector/skipper_collector.go
+++ b/pkg/collector/skipper_collector.go
@@ -88,16 +88,16 @@ func NewSkipperCollector(client kubernetes.Interface, plugin CollectorPlugin, hp
 	}, nil
 }
 
-func getAnnotationWeight(backendWeights string, backend string) float64 {
-	var weightsMap map[string]int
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #192, opened 2020-08-13
Support bracket notation in jsonPath

### LINKED ISSUE #91: [Metric-Config] JsonPath support string indexers for json prop keys with dots
## Expected Behavior
When setting up metric configs using jsonpath, it's expected to support the regular jsonpath expression.

## Actual Behavior
The actual behaviour is in the case of using string indexers for accessing json props with `.`'s in them, the jsonpath lib is not able to extract the value and go into the value.

https://github.com/zalando-incubator/kube-metrics-adapter/blob/master/pkg/collector/json_path_collector.go#L13

  1. Setup a json-path/json-key value with something like: `$.histograms['service.pool.Usage']mean`
  2. On k8s the autoscaler will report: `Failed to create new metrics collector: format '' not supported`

## Specifications

  - Version: latest

mikkeloscar:
This changes the jsonPath library used internally to a new one (https://github.com/spyzhov/ajson) which supports bracket notation for `jsonPath`. e.g. you can now ALSO use something like `$['my.metric.key']`. Before the metric key could not have dots inside as only the dot notation was supported.

Fix #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ example the following JSON data would be expected:
 ```
 
 The json-path query support depends on the
-[github.com/oliveagle/jsonpath](https://github.com/oliveagle/jsonpath) library.
+[github.com/spyzhov/ajson](https://github.com/spyzhov/ajson) library.
 See the README for possible queries. It's expected that the metric you query
 returns something that can be turned into a `float64`.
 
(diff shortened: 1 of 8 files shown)


mikkeloscar:
:+1:

aryszka:
:+1:

PR #209, opened 2020-10-01
Fix handling of invalid/empty json response

mikkeloscar:
Follow up to #192 fix handling of a json response where the jsonPath doesn't result in an Array or Number. I.e. the jsonPath is misconfigured or the response is unexpected format.

Prevents the controller from crashing with:

```
panic: runtime error: index out of range [0] with length 0
```


mikkeloscar:
:+1:

szuecs:
:+1:

PR #219, opened 2020-10-30
Use labels for mapping metric types to metrics

### LINKED ISSUE #104: Hard to distinguish which metrics query is in play ... maybe this is a feature than a defect
## Expected Behavior
 I have multiple external metrics in play similar to below,

```
    metric-config.object.istio-requests-error-rate.prometheus/query: |
      sum(rate(istio_requests_total{destination_workload="go-demo-7-primary",
               destination_workload_namespace="go-demo-7", reporter="destination",response_code=~"5.*"}[1m])) 
      / 
      sum(rate(istio_requests_total{destination_workload="go-demo-7-primary", 
               destination_workload_namespace="go-demo-7",reporter="destination"}[1m]) > 0)* 100
      or
      sum(rate(istio_requests_total{destination_workload="go-demo-7-primary", 
               destination_workload_namespace="go-demo-7",reporter="destination"}[1m])) > bool 0 * 100

    metric-config.external.prometheus-query.prometheus/istio-requests-per-replica: |
      sum(rate(istio_requests_total{destination_service_name="go-demo-7",destination_workload_namespace="go-demo-7",
                reporter="destination"}[1m])) 
      /
      count(count(container_memory_usage_bytes{namespace="go-demo-7",pod_name=~"go-demo-7-primary.*"}) by (pod_name))
    metric-config.external.prometheus-query.prometheus/istio-requests-average-resp-time: |
      sum(rate(istio_request_duration_seconds_sum{destination_workload="go-demo-7-primary", reporter="destination"}[1m])) 
      / 
      sum(rate(istio_request_duration_seconds_count{destination_workload="go-demo-7-primary", reporter="destination"}[1m]) > 0)
      or
      sum(rate(istio_request_duration_seconds_count{destination_workload="go-demo-7-primary", reporter="destination"}[1m])) 
      > bool 0

spec:
  maxReplicas: 10
  minReplicas: 1
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: go-demo-7-primary
  metrics:
  - type: External
    external:
      metric:
        name: prometheus-query
        selector:
          matchLabels:
            query-name: istio-requests-per-replica
      target:
        type: AverageValue
        value: 5
  - type: External
    external:
      metric:
        name: prometheus-query
        selector:
          matchLabels:
            query-name: istio-requests-average-resp-time
      target:
        type: Value
        value: 100m
  - type: Object
    object:
      metric:
        name: istio-requests-error-rate
      describedObject:
        apiVersion: v1 #make sure you check the api version on the targeted resource using get command.
        kind: Pod # note Pod can be used as resource kind for kube-metrics-adapter.
        name: go-demo-7-primary
      target:
        type: Value
        value: 5
```

Then when I describe HPA I would expect that I will be able to distinguish each metrics that are in play as below,
```
  "istio-requests-per-replica" (target value):                                    0 / 5
  "istio-requests-average-resp-time" (target value):                                    0 / 100m
  "istio-requests-error-rate" on Pod/go-demo-7-primary (target value):  0 / 5
```

## Actual Behavior

I see below in HPA which is very hard to distinguish which external metrics is in play
```
Metrics:                                                                ( current / target )
  "prometheus-query" (target value):                                    0 / 5
  "prometheus-query" (target value):                                    0 / 100m
  "istio-requests-error-rate" on Pod/go-demo-7-primary (target value):  0 / 5
```

  1.Install Kube-metrics-adaptor v0.1.1
  1.Create an hpa with the above metrics
  1.Describe HPA

## Specifications

  - Version:
```
    --set image.repository=registry.opensource.zalan.do/teapot/kube-metrics-adapter \
    --set image.tag=v0.1.0

```
  - Platform:
AWS KOPS
  - Subsystem:

mikkeloscar:
For historical reasons (i.e. use of previous autoscaling API versions) we have been mapping **metric names** to the different collector types. I.e. if you wanted a metric based on `prometheus` you would name the metric `prometheus-query`.
This design has the downside that multiple metrics will have the same name and it becomes hard to distinguish as reported in #104 
In some cases it can even prevent having more than a single metric of a certain type i.e. if you use the `zmon-check` metric and have the same `check-id` but different complex keys defined (complex keys must be defined as annotation, and this will overlap if you have more than a single zmon-check metric defined).

It was initially done because we didn't have any other identifiers in the HPA spec, at least for custom metrics, and this limitation was accidentally carried over for External metrics even though they have labels which can be used for such extra information. With `v2beta2` we have selector/labels for all the metric types and this can then be used to store this information.

This PR introduces support for multiple metrics of the same type by adding a convention of a `type` label on the metrics e.g. `type: prometheus` which is used to identify which collector type should be used.

I have also aligned some of the naming for the types so they are more simple i.e. `prometheus-query` becomes type `prometheus`.

To provide backwards compatibility the old metric names are still supported, but marked as *legacy* so we eventually can deprecate and remove them.

Fix #104

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ kind: HorizontalPodAutoscaler
 metadata:
   name: myapp-hpa
   annotations:
-    # metric-config.<metricType>.<metricName>.<collectorName>/<configKey>
+    # metric-config.<metricType>.<metricName>.<collectorType>/<configKey>
     metric-config.pods.requests-per-second.json-path/json-key: "$.http_server.rps"
     metric-config.pods.requests-per-second.json-path/path: /metrics
     metric-config.pods.requests-per-second.json-path/port: "9090"
@@ -104,7 +104,7 @@ kind: HorizontalPodAutoscaler
 metadata:
   name: myapp-hpa
   annotations:
-    # metric-config.<metricType>.<metricName>.<collectorName>/<configKey>
+    # metric-config.<metricType>.<metricName>.<collectorType>/<configKey>
     metric-config.pods.requests-per-second.json-path/json-key: "$.http_server.rps"
     metric-config.pods.requests-per-second.json-path/path: /metrics
     metric-config.pods.requests-per-second.json-path/port: "9090"
@@ -216,11 +216,10 @@ Default is 1 minute.
 
 This is an example of an HPA configured to get metrics based on a Prometheus
 query. The query is defined in the annotation
-`metric-config.external.prometheus-query.prometheus/processed-events-per-second`
+`metric-config.external.processed-events-per-second.prometheus/query`
 where `processed-events-per-second` is the query name which will be associated
-with the result of the query. A matching `query-name` label must be defined in
-the `matchLabels` of the metric definition. This allows having multiple
-prometheus queries associated with a single HPA.
+with the result of the query.
+This allows having multiple prometheus queries associated with a single HPA.
 
 ```yaml
 apiVersion: autoscaling/v2beta2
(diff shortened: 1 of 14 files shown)


mikkeloscar:
@szuecs added a warning log

szuecs:
:+1:

mikkeloscar:
Found a bug where the fallback didn't work in case some labels were defined for the metric. Fixed and added tests to cover this.

mikkeloscar:
:+1:

csenol:
:+1:

szuecs:
:+1:

Issue #56, opened 2019-05-09
Helm Chart


agolomoodysaada:
We should have a published helm chart for this project.

mikkeloscar:
Contributions are very welcome for this! :)

agolomoodysaada:
Happy to contribute the chart I have but blocked by: https://github.com/zalando-incubator/kube-metrics-adapter/issues/55

mikkeloscar:
Fwiw there's a chart here: https://github.com/banzaicloud/banzai-charts/tree/master/kube-metrics-adapter

balamuru:
For external metrics,  the banzai cloud chart is geared towards external / aws / sqs - i had better luck just deploying the yaml specs provided in the kube-metrics-adaptor . I haven't had a chance to go back and compare what the difference is.

Freyert:
`aws.enable` turns on the External metrics adapter, but it also turns on the AWS collector. 🤦‍♂ 
I get the feeling this chart will not meet expectations.
And the README doesn't match the actual values in some small cases, but I find it still kind of works.

mikkeloscar:
The image in the chart linked above is a rebuild/republish of our image, so it will most likely be outdated. We are not users of helm so don't have the biggest incentive to work on this and maintain it, but contributions are welcome and I'm happy to review work towards a better chart :)

PR #238, opened 2020-12-19
Add a kubernetes helm chart

tanersener:
# Adds a kubernetes helm chart

> Issue : #56

## Description
Creates a helm chart for this project that can be installed using `helm install`.

## Types of Changes
- Refactor/improvements

## Features
- Existing helm deployment files are preserved
- Helm chart is added under the `docs/helm` folder
    - Service account definition in `rbac.yaml` file is extracted as a separate `service-account.yaml` file
    - Missing `kube-metrics-adapter-resource-reader` cluster role binding is added
    - Definition names are updated with `kube-metrics-adapter` prefix
- Can be installed using `helm install`
- All docker image options except `--help` and `--bind-address` can be customized during `helm install`
- Allows changing the namespace
- Can be used to install a different docker image

## Deployment Notes
You can deploy the chart using the following command

`helm install kube-metrics-adapter helm --set prometheus.server=http://iamserver --set aws.region=eu-west1 --set prometheus.metricsAddress=:9090 --set zmon.tokenName=QTCuP5lwSQj6hdUrnGqAcYyspGY17U --set influxDB.address=http://influx --set log.level=2`


mikkeloscar:
@tanersener Thanks for this! As we don't use helm ourselves we can support this as best effort, but I would prefer if we also keep the existing non-helm resources in place. Could you change the PR to have both helm and non-helm manifests?

tanersener:
@mikkeloscar Moved the `helm` folder under the `docs`. Now we have both manifests side by side.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

tanersener:
Thanks :+1:

PR #250, opened 2021-01-05
Switch to compliant base image

jonathanbeber:
Use compliant base image according to our tech rules of play:
- KDP-1: use an approved Docker base image and version from the list of
    allowed Docker Base Images under the pierone namespace ‘library’.


mikkeloscar:
:+1:

jonathanbeber:
:+1:

Issue #271, opened 2021-02-19
no subjectaccessreviews access


victor23d:
It works on GKE cluster but doesn't work on AKS cluster.
Error log on AKS cluster:
```subjectaccessreviews.authorization.k8s.io is forbidden: User \"system:serviceaccount:custom-metrics:custom-metrics-apiserver\" cannot create resource \"subjectaccessreviews\" in API group \"authorization.k8s.io\" at the cluster scope") has prevented the request from succeeding (get pods.custom.metrics.k8s.io *)```
It's caused by lacking subjectaccessreviews access.
I could make a pull request to fix it.

**Comment by mikkeloscar** [comment]
Could you add some more logs to show where the `kube-metrics-adapter` is failing with this error? I don't understand when it needs those permissions.

**Comment by ymmt2005** [comment]
There should be a system ClusterRole named `system:auth-delegator` for granting access to SubjectAccessReviews API.
ref. https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/#extension-apiserver-authorizes-the-request

As the current manifest has ClusterRoleBinding for `system:auth-delegator` already, I don't understand the reason too.
Maybe because AKS clusters don't have `system:auth-delegator` ClusterRole?

**Comment by szuecs** [comment]
@ymmt2005 interesting, maybe @victor23d can check that. In our custom AWS cluster (non eks) we have the clusterrole system:auth-delegator

**Comment by szuecs** [comment]
I got the answer:

```
$ kubectl get clusterrole -o yaml system:auth-delegator
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  creationTimestamp: "2021-02-27T08:43:42Z"
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
rules:
- apiGroups:
  - authentication.k8s.io
  resources:
  - tokenreviews
  verbs:
  - create
- apiGroups:
  - authorization.k8s.io
  resources:
  - subjectaccessreviews
  verbs:
  - create
```
from a 1.19.7 cluster

PR #272, opened 2021-02-19
add subjectaccessreviewer in rbac.yaml

victor23d:
Fix no subjectaccessreviews access #271 
> Issue : #271


mikkeloscar:
Closing this for now as there is still unresolved questions in #271 to clarify why the change is needed. We are happy to re-consider once we have some clarity on the problem in #271

PR #284, opened 2021-03-19
Create first version of ScalingSchedule CRD

jonathanbeber:
Some cases and users actually know when their applications will have a
high workload and, therefore, autoscaling support for time based scaling
is a desired feature.

This commit creates the first version of two CRDs called
`ScalingSchedule` and `ClusterScalingSchedule`.
The CRDs describe one or multiples schedules inside them. The schedules
contains the information of when the time based scaling starts, if it
happens once or multiple times, its duration and, a configurable value
that later can be used by HPAs to scale applications.

The only difference between the two CRDs is their scope.
`ClusterScalingSchedule` aims to attend cluster wide schedules, to
multiple applications, while `ScalingSchedule` has to be deployed with
each application into the same namespace.

This commit does not create any metric, it's a noop change that creates
just the CRD and import tools required to generate the CRD and others
required code, as `deepCopy` functions and clients.


jonathanbeber:
It's related also to https://github.com/kubernetes/kubernetes/issues/50569

docs/scaling_schedules_crd.yaml
```diff
@@ -0,0 +1,118 @@
...
+  creationTimestamp: null
+  name: scalingschedules.zalando.org
+spec:
+  group: zalando.org
+  names:
+    kind: ScalingSchedule
+    listKind: ScalingScheduleList
+    plural: scalingschedules
+    singular: scalingschedule
+  scope: Namespaced
```
jonathanbeber, line 17: @mikkeloscar regarding our last conversation: after a closer look my idea is to go with just one namespace CRD and make it possible for HPAs to define the object namespace if not the default. The default isn't `default`, I would consider, in the implementation, the same namespace as the HPA itself. Just like name resolution for services. What do you think? Is it feasible? Do you see drawbacks?

mikkeloscar: The problem here is that you can't specify the `namespace` in the HPA object reference: https://github.com/kubernetes/kubernetes/blob/cf3374e43491c594071548f68d4357fc9ed537ea/staging/src/k8s.io/api/autoscaling/v2beta2/types.go#L86-L94

So unless we would encode it in the name (or use an external metric type to reference it) I don't know how to get around defining both a namespaced and cluster spaced resource type if we want to support both use cases.

jonathanbeber: Thanks for the context, you're right, we will have to support both, which shouldn't be a problem. Will fix.

pkg/apis/zalando.org/v1/types.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Schedule is the schedule details to be used inside a ScalingSchedule.
+// +k8s:deepcopy-gen=true
+type Schedule struct {
+	// Defines if the schedule is a OneTime schedule or
+	// Repeating one. If OneTime, date has to be defined. If Repeating,
+	// Period has to be defined.
+	// +kubebuilder:validation:Enum=OneTime;Repeating
+	ScheduleType string `json:"type"`
```
mikkeloscar, line 53: nit: Maybe we want to make this into a type similar to here: https://github.com/zalando-incubator/stackset-controller/blob/6348bf62633d184eab0dbe59b0b5482d81bc460f/pkg/apis/zalando.org/v1/types.go#L173-L188

same for `ScheduleDay`

jonathanbeber: good one, will do.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #295, opened 2021-04-08
Bugfix/fix disregard incompatible hpas error handling

adutchak-x:
# One-line summary

> Changed error type check for disregard-incompatible-hpas option handler

## Description
Changed error type check from errors.Is() to errors.As() because errors.Is() always returns False.
In our installation where we have other external metrics adapter with own metrics this was still flooding our HPA resources with events like "type: 'Warning' reason: 'CreateNewMetricsCollector' Failed to create new metrics collector: no plugin found for {External {xxx-xxx-xxx nil}}" however we've set --disregard-incompatible-hpas key to true.

## Types of Changes


jonathanbeber:
Hello, @adutchak-x. Thanks for your PR.

This is strange we have to change the API to set the err as another type to verify its type. I guess it tells a bit about error handling in golang :confused: 

I believe the best way to handle that would be to implement the barely documented interface `Is(err) bool`: https://go.googlesource.com/go/+/go1.16.2/src/errors/wrap.go#30

What do you think?

adutchak-x:
Hello @jonathanbeber,
Agree, following your suggestion, I've implemented Is() method for PluginNotFoundError, it makes only type assertion, not sure what more we can check within it.

pkg/collector/collector.go
```diff
@@ -61,6 +61,12 @@ func (p *PluginNotFoundError) Error() string {
 	return fmt.Sprintf("no plugin found for %s", p.metricTypeName)
 }
 
+func (p *PluginNotFoundError) Is(target error) bool {
+	t, ok := target.(*PluginNotFoundError)
+	_ = t
```
jonathanbeber, line 66:
```suggestion
	_, ok := target.(*PluginNotFoundError)
```

quick :nail_care:, the rest it looks good to me.

jonathanbeber:
@mikkeloscar any toughts?

mikkeloscar:
hmm, I actually think the original idea of switching to `errors.As` is better because that is intended to compare the type as opposed to the value. Now we implement `Is` and do type assertion there, basically implementing `As`. :) - but these are details, I'm also ok with how it is now if you prefer.

However, what is more important is to add a test case since this was not covered in the original implementation of this error handling. @adutchak-x do you mind adding a test case for this, so we can validate that it works as expected!

jonathanbeber:
I see, yes, I'm also fine with both solutions. I think in any case it's strange, but that's what we have. What I don't like about the `.As` solution is that we have to create the object to receive it anyway and we clearly do anything with the `var pluginNotFoundError *collector.PluginNotFoundError`, but ofc, that's something I can live with.

I agree, we need to have a test for that. @adutchak-x if you need help with this one just let me know.

adutchak-x:
> I agree, we need to have a test for that. @adutchak-x if you need help with this one just let me know.

There is actually a test for that already - https://github.com/zalando-incubator/kube-metrics-adapter/blob/master/pkg/provider/hpa_test.go#L98 but I think it does not work properly since the issue appeared in a master branch.
I'll try to sort it out.

adutchak-x:
@jonathanbeber @mikkeloscar,
so the issue with _**TestUpdateHPAsDisregardingIncompatibleHPA**_  was that **_updateHPAs()_** method was always returning "nil" except failed attempt to list the HPAs at the very beginning. This led to having the test always passing.

I've just updated the code so that we collect errors inside the loops and give them to process the rest of the stuff, once this work is completed we return an array of errors instead of "nil" or actually "nil" if no errors happened.

jonathanbeber:
:+1:

mikkeloscar:
@adutchak-x thanks for your effort on this. Regarding the changes I'm now a little concerned that we change the error handling to handle the errors twice (first log them and then return them from `updateHPAs()` to be logged a second time). This also means that the error counter metric now have a different meaning where any invalid HPA can cause it to go up which wasn't intended, it's more intended to know if the metrics-adapter is not behaving correctly as opposed to counting the number of invalid HPAs.

What we really want to check in the test is if we get an event. We could mock the event recorder and check if we got the expected event in the test instead of checking of `err`. What do you think? I'm happy to help with the test if you like!

adutchak-x:
> @adutchak-x thanks for your effort on this. Regarding the changes I'm now a little concerned that we change the error handling to handle the errors twice (first log them and then return them from `updateHPAs()` to be logged a second time). This also means that the error counter metric now have a different meaning where any invalid HPA can cause it to go up which wasn't intended, it's more intended to know if the metrics-adapter is not behaving correctly as opposed to counting the number of invalid HPAs.
> 
> What we really want to check in the test is if we get an event. We could mock the event recorder and check if we got the expected event in the test instead of checking of `err`. What do you think? I'm happy to help with the test if you like!

@mikkeloscar ok, I was thinking about the same, just didn't understand the meaning of the metrics collected in here: https://github.com/zalando-incubator/kube-metrics-adapter/blob/master/pkg/provider/hpa.go#L103
Can you help me indeed with the event recorder?

mikkeloscar:
@adutchak-x I made a PR to your PR with a proposal for how to do it: https://github.com/adutchak-x/kube-metrics-adapter/pull/1

adutchak-x:
> @adutchak-x I made a PR to your PR with a proposal for how to do it: [adutchak-x#1](https://github.com/adutchak-x/kube-metrics-adapter/pull/1)

Fantastic,
I'd not be able to fix that in that way :) (I'm a bit new to Golang)

mikkeloscar:
:+1:

mikkeloscar:
@adutchak-x Thanks for all your patience and effort on this one!

jonathanbeber:
:+1:

jonathanbeber:
thanks, @adutchak-x!

adutchak-x:
You're welcome @jonathanbeber

PR #298, opened 2021-04-13
Clearly document interval config

### LINKED ISSUE #296: Metric Resolution Flag
Hi,

I can set  the flag in the metric-server, thus I can query for the predefined metrics (CPU and memory) more quickly (10s, 30s and so on). I want to do this too for the custom and external metrics, because they are queried once in a minute. 

Is there a way to set the metric-resolution flag for the custom and external metrics?

Thank you.

mikkeloscar:
Documents the interval config clearly!

Fix #296


jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #308, opened 2021-04-16
Feature/improve pod collector

adutchak-x:
# Improved pod collector, introduced **pod-min-age** configuration option

## Description
In order to prevent the situations when fresh pods can return unstable metrics while they're warming up - introducing `min-pod-age` configuration option. This option allows us to add a delay before collector starts gathering metrics from the pods.

Also, I've noticed that pod collector starts querying pods without consideration of their status. I've added IsPodReady conditional check before attempting querying the metrics.

For example, In our case we're using an upstream response times as an additional scaling metric.
However, when we're performing the deployments, the values for response times are very high for the first couple of seconds after new pods have been started. This leads to the situation when kube-metrics-adapter triggers a scale up of our HPAs.

## Types of Changes

- Configuration change
- Documentation / non-code

## Review
- [ X] Tests
- [ X] Documentation
- [ X] CHANGELOG

### DIFF
--- a/pkg/collector/pod_collector.go
+++ b/pkg/collector/pod_collector.go
@@ -180,7 +180,7 @@ func GetPodReadyAge(pod corev1.Pod) (bool, time.Duration) {
 		return false, podReadyAge
 	}
 	for i := range conditions {
-		if conditions[i].Type == corev1.PodReady {
+		if conditions[i].Type == corev1.PodReady && conditions[i].Status == corev1.ConditionTrue {
 			podReadyAge = time.Since(conditions[i].LastTransitionTime.Time)
 			return true, podReadyAge
 		}
--- a/pkg/collector/pod_collector.go
+++ b/pkg/collector/pod_collector.go
@@ -103,11 +103,11 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
 				go c.getPodMetric(pod, ch, errCh)
 			} else {
 				skippedPodsCount++
-				c.logger.Warnf("Skipping metrics collection for pod %s because it's ready age is %s and min-pod-ready-age is set to %s", pod.Name, podReadyAge, c.minPodReadyAge)
+				c.logger.Warnf("Skipping metrics collection for pod %s/%s because it's ready age is %s and min-pod-ready-age is set to %s", pod.Namespace, pod.Name, podReadyAge, c.minPodReadyAge)
 			}
 		} else {
 			skippedPodsCount++
-			c.logger.Warnf("Skipping metrics collection for pod %s because it's status is not Ready.", pod.Name)
+			c.logger.Warnf("Skipping metrics collection for pod %s/%s because it's status is not Ready.", pod.Namespace, pod.Name)
 		}
 	}
 
@@ -172,9 +172,8 @@ func getPodLabelSelector(client kubernetes.Interface, hpa *autoscalingv2.Horizon
 }
 
 // GetPodReadyAge extracts corev1.PodReady condition from the given pod object and
-// returns true, time.Duration() for pod.LastTransitionTime if the condition corev1.PodReady is found. Returns time.Duration(0s), false if the condition is not present.
+// returns true, time.Duration() for LastTransitionTime if the condition corev1.PodReady is found. Returns time.Duration(0s), false if the condition is not present.
 func GetPodReadyAge(pod corev1.Pod) (bool, time.Duration) {
-	t := time.Now()
 	podReadyAge := time.Duration(0 * time.Second)
 	conditions := pod.Status.Conditions
 	if conditions == nil {
@@ -182,7 +181,8 @@ func GetPodReadyAge(pod corev1.Pod) (bool, time.Duration) {
 	}
 	for i := range conditions {
 		if conditions[i].Type == corev1.PodReady {
-			podReadyAge = time.Duration(t.Sub(conditions[i].LastTransitionTime.Time).Nanoseconds())
+			podReadyAge = time.Since(conditions[i].LastTransitionTime.Time)
+			fmt.Printf("podReadyAge: %s", podReadyAge)
 			return true, podReadyAge
 		}
 	}
--- a/pkg/collector/pod_collector.go
+++ b/pkg/collector/pod_collector.go
@@ -38,7 +38,7 @@ type PodCollector struct {
 	namespace        string
 	metric           autoscalingv2.MetricIdentifier
 	metricType       autoscalingv2.MetricSourceType
-	minPodAge				 time.Duration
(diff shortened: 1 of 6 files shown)


pkg/collector/pod_collector.go
```diff
@@ -152,3 +169,18 @@ func getPodLabelSelector(client kubernetes.Interface, hpa *autoscalingv2.Horizon
 
 	return nil, fmt.Errorf("unable to get pod label selector for scale target ref '%s'", hpa.Spec.ScaleTargetRef.Kind)
 }
+
+// IsPodReady extracts corev1.PodReady condition from the given pod object and
+// returns the true if the condition corev1.PodReady is found. Returns -1 and false if the condition is not present.
```
mikkeloscar, line 174: nit: it doesn't return `-1`

pkg/collector/pod_collector.go
```diff
@@ -89,12 +91,27 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
...
 	errCh := make(chan error)
+	skippedPodsCount := 0
+
 	for _, pod := range pods.Items {
-		go c.getPodMetric(pod, ch, errCh)
+		t := time.Now()
+		podAge := time.Duration(t.Sub(pod.ObjectMeta.CreationTimestamp.Time).Nanoseconds())
+
+		if podAge > c.minPodAge {
+			if IsPodReady(pod) {
```
mikkeloscar, line 101: Considering that we filter for PodReady which makes sense, I think the `min-pod-age` should actually be `min-pod-ready-age` meaning the minimum time since the pod got ready as opposed to the time since the pod was created. If you only consider the pod creation time you can in theory have a pod pending for hours and then when it gets ready at some point the `min-pod-age` will be satisfied right away even though the container may just have started.

The pod conditions have a `lastTransitionTime` field which can be used to determine how long a pod was ready.

mikkeloscar (changes requested):

pkg/collector/pod_collector.go
adutchak-x: Got it, will correct the comment.

pkg/collector/pod_collector.go
adutchak-x: Yes, this makes sense. I'll try to figure out how to calculate that time.

adutchak-x: Hello, now it takes lastTransitionTime as a target for min-pod-ready-age calculation

pkg/collector/pod_collector.go
```diff
@@ -89,12 +92,27 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
...
-		go c.getPodMetric(pod, ch, errCh)
+
+		isPodReady, podReadyAge := GetPodReadyAge(pod)
+
+		if isPodReady {
+			if podReadyAge > c.minPodReadyAge {
+				go c.getPodMetric(pod, ch, errCh)
+			} else {
+				skippedPodsCount++
+				c.logger.Warnf("Skipping metrics collection for pod %s because it's ready age is %s and min-pod-ready-age is set to %s", pod.Name, podReadyAge, c.minPodReadyAge)
```
mikkeloscar, line 106: let's log both pod name and namespace like this: `<namespace>/<name>` to make it simpler to understand in which namespace a pod is located.

Same for the log below.

pkg/collector/pod_collector.go
```diff
@@ -152,3 +170,22 @@ func getPodLabelSelector(client kubernetes.Interface, hpa *autoscalingv2.Horizon
...
+func GetPodReadyAge(pod corev1.Pod) (bool, time.Duration) {
+	t := time.Now()
+	podReadyAge := time.Duration(0 * time.Second)
+	conditions := pod.Status.Conditions
+	if conditions == nil {
+		return false, podReadyAge
+	}
+	for i := range conditions {
+		if conditions[i].Type == corev1.PodReady {
+			podReadyAge = time.Duration(t.Sub(conditions[i].LastTransitionTime.Time).Nanoseconds())
```
mikkeloscar, line 185: nit: you can simplify this to `time.Since`: https://golang.org/pkg/time/#Since

pkg/collector/pod_collector.go
adutchak-x: Fixed

pkg/collector/pod_collector.go
adutchak-x: Used time.Since

pkg/collector/pod_collector.go
```diff
@@ -89,12 +92,27 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
...
 	errCh := make(chan error)
+	skippedPodsCount := 0
+
 	for _, pod := range pods.Items {
-		go c.getPodMetric(pod, ch, errCh)
+
+		isPodReady, podReadyAge := GetPodReadyAge(pod)
+
+		if isPodReady {
+			if podReadyAge > c.minPodReadyAge {
```
mikkeloscar, line 102: nit: we probably want `>=` here (very small detail but more semantically correct) :)

pkg/collector/pod_collector.go
```diff
@@ -152,3 +170,22 @@ func getPodLabelSelector(client kubernetes.Interface, hpa *autoscalingv2.Horizon
...
+func GetPodReadyAge(pod corev1.Pod) (bool, time.Duration) {
+	podReadyAge := time.Duration(0 * time.Second)
+	conditions := pod.Status.Conditions
+	if conditions == nil {
+		return false, podReadyAge
+	}
+	for i := range conditions {
+		if conditions[i].Type == corev1.PodReady {
+			podReadyAge = time.Since(conditions[i].LastTransitionTime.Time)
+			fmt.Printf("podReadyAge: %s", podReadyAge)
```
mikkeloscar, line 185: drop the debug print from there :)

mikkeloscar:
@adutchak-x The last few nits from my side and then it looks good!

pkg/collector/pod_collector.go
adutchak-x: :) Sorry, indeed

mikkeloscar:
:+1:

mikkeloscar (approved):

adutchak-x:
@mikkeloscar Fortunately noticed in our dev environment that was not really verifying the Ready condition without additional check like this: `&& conditions[i].Status == corev1.ConditionTrue.`Just pushed the fix.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

jonathanbeber:
thanks @adutchak-x

PR #310, opened 2021-04-26
Updated apiservice to v1

Miniland1333:
# Updated APIService from v1beta1 to v1

## Description
When installing this adapter into Kubernetes v1.19+, you will receive the following deprecation notice:
```
W0426 10:49:35.932780     318 warnings.go:70] apiregistration.k8s.io/v1beta1 APIService is deprecated in v1.19+, unavailable in v1.22+; use apiregistration.k8s.io/v1 APIService
```

This migration was described by the [KubernetesDeprecation Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#apiservice-v122) as one with "No notable changes", so there should be no functionality change.

## Types of Changes

## Tasks
  - [x] Replaced all instances of `apiregistration.k8s.io/v1beta1` with `apiregistration.k8s.io/v1`

## Review
- [ ] Verify that `apiregistration.k8s.io/v1` substitutions are correct and not misspelled.

## Deployment Notes
Should be deployed before Kubernetes v1.22.

### DIFF
--- a/docs/custom-metrics-apiservice.yaml
+++ b/docs/custom-metrics-apiservice.yaml
@@ -1,13 +1,13 @@
 apiVersion: apiregistration.k8s.io/v1
 kind: APIService
 metadata:
-  name: v1.custom.metrics.k8s.io
+  name: v1beta1.custom.metrics.k8s.io
 spec:
   service:
     name: kube-metrics-adapter
     namespace: kube-system
   group: custom.metrics.k8s.io
-  version: v1
+  version: v1beta1
   insecureSkipTLSVerify: true
   groupPriorityMinimum: 100
   versionPriority: 100
--- a/docs/custom-metrics-apiservice.yaml
+++ b/docs/custom-metrics-apiservice.yaml
@@ -1,13 +1,13 @@
-apiVersion: apiregistration.k8s.io/v1beta1
+apiVersion: apiregistration.k8s.io/v1
 kind: APIService
 metadata:
-  name: v1beta1.custom.metrics.k8s.io
+  name: v1.custom.metrics.k8s.io
 spec:
   service:
     name: kube-metrics-adapter
     namespace: kube-system
   group: custom.metrics.k8s.io
-  version: v1beta1
+  version: v1
   insecureSkipTLSVerify: true
   groupPriorityMinimum: 100
   versionPriority: 100
--- a/docs/external-metrics-apiservice.yaml
+++ b/docs/external-metrics-apiservice.yaml
@@ -1,13 +1,13 @@
-apiVersion: apiregistration.k8s.io/v1beta1
+apiVersion: apiregistration.k8s.io/v1
 kind: APIService
 metadata:
-  name: v1beta1.external.metrics.k8s.io
+  name: v1.external.metrics.k8s.io
 spec:
   service:
     name: kube-metrics-adapter
     namespace: kube-system
   group: external.metrics.k8s.io
-  version: v1beta1
+  version: v1
   insecureSkipTLSVerify: true
   groupPriorityMinimum: 100
   versionPriority: 100
--- a/docs/external-metrics-apiservice.yaml
+++ b/docs/external-metrics-apiservice.yaml
@@ -1,13 +1,13 @@
 apiVersion: apiregistration.k8s.io/v1
 kind: APIService
 metadata:
-  name: v1.external.metrics.k8s.io
+  name: v1beta1.external.metrics.k8s.io
 spec:
   service:
     name: kube-metrics-adapter
     namespace: kube-system
   group: external.metrics.k8s.io
-  version: v1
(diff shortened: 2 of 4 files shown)


jonathanbeber:
:+1:

docs/custom-metrics-apiservice.yaml
```diff
@@ -1,13 +1,13 @@
-apiVersion: apiregistration.k8s.io/v1beta1
+apiVersion: apiregistration.k8s.io/v1
 kind: APIService
 metadata:
-  name: v1beta1.custom.metrics.k8s.io
```
jonathanbeber, line 4: @mikkeloscar you know if there's any reason to use [`v1beta1.custom.metrics.k8s.io`](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/kube-metrics-adapter/custom-metrics-apiservice.yaml#L4)?

docs/custom-metrics-apiservice.yaml
```diff
@@ -1,13 +1,13 @@
...
+  name: v1.custom.metrics.k8s.io
 spec:
   service:
     name: kube-metrics-adapter
     namespace: kube-system
   group: custom.metrics.k8s.io
-  version: v1beta1
```
jonathanbeber, line 10: same [here](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/kube-metrics-adapter/custom-metrics-apiservice.yaml#L10)

docs/custom-metrics-apiservice.yaml
jonathanbeber: Ok, so far as I could check, that's just the name and we don't refer to that anywhere else.

docs/custom-metrics-apiservice.yaml
jonathanbeber: This one theoretically defines the version the server talks. Since there's no notable change and we don't strictly define it, I think it's ok too. I'd be interested in updating it on our environment too.

jonathanbeber:
@Miniland1333 do you mind signing off the commits?

Miniland1333:
> @Miniland1333 do you mind signing off the commits?

Do you know if there is a way to do this through the Github interface for those commits? I just edited them online it Github which is why the commits are <verified>.

jonathanbeber:
I don't think it's possible, unfortunately, but I'm not 100% sure.

docs/custom-metrics-apiservice.yaml
Miniland1333: At least in my local testing today I am using v1 and I haven't encountered any issues so far. I'll let you know if anything pops up.

mikkeloscar: I think it maps to the versions defined here: https://github.com/kubernetes/metrics/tree/master/pkg/apis/custom_metrics so we should not update this one otherwise it exposes an unexpected API as v1 doesn't exist.

jonathanbeber: makes more sense now. So we can follow [these examples](https://github.com/kubernetes-sigs/custom-metrics-apiserver/blob/0ca2b1909cdc45259654a4e06cac40f98fa1bcb0/test-adapter-deploy/testing-adapter.yaml#L108-L134)  and update just the API version of the `APIService` itself. @Miniland1333 do you mind reverting these changes?

docs/custom-metrics-apiservice.yaml
jonathanbeber: same as [above](https://github.com/zalando-incubator/kube-metrics-adapter/pull/310#discussion_r620968482).

docs/custom-metrics-apiservice.yaml
Miniland1333: That makes sense and it explains the weird issue I was experiencing at the end of yesterday. Also, quick local testing indicates that metrics API v1beta2 is not a drop-in replacement as when querying the endpoint you get `Error from server (ServiceUnavailable): the server is currently unable to handle the request`
image

Would you like me to add the blocks for `v1beta2.external.metrics.k8s.io` as you suggested in [this example](https://github.com/kubernetes-sigs/custom-metrics-apiserver/blob/0ca2b1909cdc45259654a4e06cac40f98fa1bcb0/test-adapter-deploy/testing-adapter.yaml#L108-L134) or since it isn't currently working do you instead want it to be in a separate PR.

mikkeloscar: Let's do this separately.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

jonathanbeber:
thank you @Miniland1333

Miniland1333:
Thanks for your help with this PR and for making this plugin!

PR #312, opened 2021-04-30
Feature/stop processing non runing pods

adutchak-x:
Signed-off-by: Anatolii Dutchak <adutchak-x@tunein.com>

# One-line summary

Skipping pods which might be in "Termination" state

## Description
It seems to be that we did not consider the situation where some of the pods can be in "Terminating" state. We actually don't want to collect the metrics from pods scheduled for deletion.

## Types of Changes

- Refactor/improvements

## Review
- [X ] Tests

### DIFF
--- a/pkg/collector/pod_collector.go
+++ b/pkg/collector/pod_collector.go
@@ -101,7 +101,7 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
 		if isPodReady {
 			if pod.DeletionTimestamp != nil {
 				skippedPodsCount++
-				c.logger.Warnf("Skipping metrics collection for pod %s/%s because it is seems to be scheduled for termination (DeletionTimestamp: %s)", pod.Namespace, pod.Name, pod.DeletionTimestamp)
+				c.logger.Debugf("Skipping metrics collection for pod %s/%s because it is being terminated (DeletionTimestamp: %s)", pod.Namespace, pod.Name, pod.DeletionTimestamp)
 			} else if podReadyAge < c.minPodReadyAge {
 				skippedPodsCount++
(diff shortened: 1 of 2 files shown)


pkg/collector/pod_collector.go
```diff
@@ -99,11 +99,14 @@ func (c *PodCollector) GetMetrics() ([]CollectedMetric, error) {
 		isPodReady, podReadyAge := GetPodReadyAge(pod)
 
 		if isPodReady {
-			if podReadyAge >= c.minPodReadyAge {
-				go c.getPodMetric(pod, ch, errCh)
-			} else {
+			if pod.DeletionTimestamp != nil {
+				skippedPodsCount++
+				c.logger.Warnf("Skipping metrics collection for pod %s/%s because it is seems to be scheduled for termination (DeletionTimestamp: %s)", pod.Namespace, pod.Name, pod.DeletionTimestamp)
```
mikkeloscar, line 104: nit: I would make this message more clear and say: `Skipping metrics collection for pod %s/%s because it is being terminated (DeletionTimestamp: %s)`

I would also propose we just do this a Debug log to be less noisy.

adutchak-x: @mikkeloscar okay, agree about the message. Should set a debug level only for this kind of message of for other two too?
Maybe we can leave Warn for "podReadyAge < c.minPodReadyAge" condition check and "pod not ready ready" & "pod terminating" cases can go to the debug level?

mikkeloscar: Sounds good to me!

adutchak-x: Pushed.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

jonathanbeber:
thanks @adutchak-x

PR #315, opened 2021-05-21
Create ScalingSchedule collector

jonathanbeber:
# One-line summary

Create `ScalingSchedule` and `ClusterScalingSchedule` collectors.

## Description
This commit adds two new collectors to the adapter:
- ClusterScalingScheduleCollector; and
- ScalingScheduleCollector

Also, it introduces the required collectors plugins, initialization
logic in the server startup, documentation and deployment example
(including the helm chart). A new config flag is created,
`-scaling-schedule`, and allows to enable and to disable the collection
of such metrics. It's disabled by default.

This collectors are the required logic to utilise the CRDs introduced in
the #284 pull request. It makes use of the Kubernetes go-client
implementations of a [Store][0] and [Reflector][1].

[0]: https://pkg.go.dev/k8s.io/client-go/tools/cache#Store
[1]: https://pkg.go.dev/k8s.io/client-go/tools/cache#Reflector

## Types of Changes

## Review
- [ ] Tests
- [x] Documentation
- [x] CHANGELOG

## Deployment Notes
None by default, the feature is disabled by default.


jonathanbeber:
It's related also to https://github.com/kubernetes/kubernetes/issues/50569 and https://github.com/kubernetes/kubernetes/issues/49931

pkg/collector/scaling_schedule_collector.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+// GetMetrics is the main implementation for collector.Collector interface
+func (c *ScalingScheduleCollector) GetMetrics() ([]CollectedMetric, error) {
+	scalingScheduleInterface, exists, err := c.store.GetByKey(fmt.Sprintf("%s/%s", c.objectReference.Namespace, c.objectReference.Name))
+	if !exists {
+		return nil, ErrScalingScheduleNotFound
+	}
+	if err != nil {
+		return nil, fmt.Errorf("Unexpected error retrieving the ScalingSchedule: %s", err.Error())
```
mikkeloscar, line 179: nit: error text should start lowercase so they can be chained

pkg/collector/scaling_schedule_collector.go
```diff
@@ -0,0 +1,301 @@
...
+				location, err = time.LoadLocation(defaultTimeZone)
+				if err != nil {
+					return nil, fmt.Errorf("unexpected error loading default location: %s", err.Error())
+				}
+			}
+			nowInLocation := now.In(location)
+			weekday := nowInLocation.Weekday()
+			for _, day := range schedule.Period.Days {
+				if days[day] == weekday {
+					parsedStartTime, _ := time.Parse(hourColonMinuteLayout, string(schedule.Period.StartTime))
```
mikkeloscar, line 248: nit: Isn't `schedule.Period.StartTime` already a `string`? And shouldn't we handle the potential error here (or comment why we don't because of CRD validation).

README.md
```diff
@@ -671,3 +671,91 @@ metric-config.<metricType>.<metricName>.<collectorType>/interval: "30s"
...
+
+The `ScalingSchedule` and `ClusterScalingSchedule` collectors allow
+collecting time-based metrics from the respective CRD objects specified
+in the HPA.
+
+### Supported metrics
+
+| Metric | Description | Type | K8s Versions |
+| ---------- | -------------- | ------- | -- |
+| ObjectName | The metric is calculated and stored for each `ScalingSchedule` and `ClusterScalingSchedule` referenced in the HPAs | `ScalingSchedule` and `ClusterScalingSchedule` | `>=1.12` |
```
mikkeloscar, line 685: This should probably be `>=1.16` since the CRD is `apiextensions.k8s.io/v1` which is only supported since `1.16`.

README.md
```diff
@@ -671,3 +671,91 @@ metric-config.<metricType>.<metricName>.<collectorType>/interval: "30s"
...
+This resource defines a scheduling event named `scheduling-event` with
+two schedules of the kind `ClusterScalingSchedule`.
+
+`ClusterScalingSchedule` objects aren't namespaced, what means it can be
+referenced by any HPA in any namespace in the cluster. `ScalingSchedule`
+have the exact same fields and behavior, but can be referenced just by
+HPAs in the same namespace. The schedules can have the type `Repeating`
+or `OneTime`.
+
+Check the CRDs definitions in the `docs` directory for a better
```
mikkeloscar, line 724: nit: Maybe link directly to them :)

README.md
```diff
@@ -671,3 +671,91 @@ metric-config.<metricType>.<metricName>.<collectorType>/interval: "30s"
...
+    object:
+      describedObject:
+        apiVersion: zalando.org/v1
+        kind: ClusterScalingSchedule
+        name: "scheduling-event"
+      metric:
+        name: "scheduling-event"
+      target:
+        type: AverageValue
+        averageValue: "1"
```
mikkeloscar, line 753: We should probably explain that in this example it would mean that it scales to `12` in the period defined, just to make it obvious what the relation is.

pkg/server/start.go
```diff
@@ -118,6 +123,8 @@ func NewCommandStartAdapterServer(stopCh <-chan struct{}) *cobra.Command {
 		"disregard failing to create collectors for incompatible HPAs")
 	flags.DurationVar(&o.MetricsTTL, "metrics-ttl", 15*time.Minute, "TTL for metrics that are stored in in-memory cache.")
 	flags.DurationVar(&o.GCInterval, "garbage-collector-interval", 10*time.Minute, "Interval to clean up metrics that are stored in in-memory cache.")
+	flags.BoolVar(&o.ScalingScheduleMetrics, "scaling-schedule", o.ScalingScheduleMetrics, ""+
+		"whether to enable time-based ScalingSchedule metrics")
```
mikkeloscar, line 127: Alternatively we could enable it only if the CRD is available. I think this is fine for now, but there may be users who doesn't want to use e.g. `ClusterScalingSchedule` and they could control it by not deploying the CRD for instance.

Just an idea, doesn't have to be done now at all.

mikkeloscar (commented):
Looks really great! I only added a few nitpicks.

Didn't you need to add some annotations to `types.go` for correctly generating the CRD clients or was this not needed after all? (e.g. `+genclient:nonNamespaced`)

jonathanbeber:
Thanks a lot for the quick review, @mikkeloscar. 

> Didn't you need to add some annotations to `types.go` for correctly generating the CRD clients or was this not needed after all? (e.g. `+genclient:nonNamespaced`)

Not for us, but if someone else wants to use the clients, it would be "broke". I have a branch with that ready, just didn't want to make this PR even bigger since it won't change anything. In this PR we use the REST client.

pkg/server/start.go
jonathanbeber: It's a good one, I will add a note to do that in another PR.

jonathanbeber: #316 so I don't forget.

README.md
jonathanbeber: Yeah, makes sense, a good catch.

README.md
jonathanbeber: I was not sure about this one, thanks a lot for checking.

pkg/collector/scaling_schedule_collector.go
jonathanbeber: it's indeed a string, I'm not sure why I tried to convert it.

Regarding the error, you're completely right, I forgot about it, will push a change and test validating it.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

pkg/collector/scaling_schedule_collector.go
```diff
@@ -0,0 +1,309 @@
...
+	for _, schedule := range schedules {
+		switch schedule.Type {
+		case v1.RepeatingSchedule:
+			location, err := time.LoadLocation(schedule.Period.Timezone)
+			if schedule.Period.Timezone == "" || err != nil {
+				location, err = time.LoadLocation(defaultTimeZone)
+				if err != nil {
+					return nil, fmt.Errorf("unexpected error loading default location: %s", err.Error())
+				}
+			}
```
AlexanderYastrebov, line 248: If I read this correctly it would silently fallback to `defaultTimezone` in case of invalid timezone (e.g. typo) which could be a surprise for the user.

The `if schedule.Period.Timezone == ""` check does not make much sense as `time.LoadLocation` would return `UTC` for the empty timezone according to the docs https://pkg.go.dev/time#LoadLocation. Besides that I think default timezone (if still used) should be `UTC` as it is common for other timestamps in k8s.

IMO timezone should be either mandatory or empty by default which would result in UTC timezone and the logic should fail fast on error without any fallback unless I am missing something else here.

AlexanderYastrebov: Timezone is required by CRD

JCMais:
how to use the helm chart added here?

PR #325, opened 2021-05-21
Remove ClusterScalingSchedule clients namespace

jonathanbeber:
During #284 we set the ClusterScalingSchedule scope as `Cluster` but
just for kubebuilder. This flag has no effect in the generated clients.
It means the generated clients expect a namespace argument when handling
ClusterScalingSchedule resources. It works fine to list all of the
resources but makes it impossible to request a specific resource
defining its name.

This commit sets the `+genclient:nonNamespaced` flag for the CRD and
adds the regenerated clients. It does not affect any of the usages of the
clients in this codebase so far.

Also discussed [here](https://github.com/zalando-incubator/kube-metrics-adapter/pull/315#issuecomment-845825438).

### DIFF
--- a/pkg/apis/zalando.org/v1/types.go
+++ b/pkg/apis/zalando.org/v1/types.go
@@ -19,6 +19,7 @@ type ScalingSchedule struct {
 }
 
 // +genclient
+// +genclient:nonNamespaced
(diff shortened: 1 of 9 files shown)


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #326, opened 2021-05-25
Ignore ContainerResource metrics

mikkeloscar:
`ContainerResource` type metrics are supported as alpha feature in Kubernetes v1.20. They are implemented in the in-tree HPA, so we don't need/want to support them here similar to how we don't support `Resource` type metrics.

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#container-resource-metrics


mikkeloscar:
:+1:

PR #334, opened 2021-06-30
Switch Schedule optional fields to pointers

jonathanbeber:
The `Date` and `Period` fields inside the `Schedule` type of the
`[Cluster]ScalingSchedule` CRDs are optional, but in its current
configuration the generated clients can't update or create these
resources since the empty fields do not pass the validation.

This commit updates the `[Cluster]ScalingSchedule.Schedule[*]` `Date`
and `Period` field to pointers, this way a null value is not validated
and the clients are able to update and create resources. It also updates
the collector code and tests to reflect the change.


jonathanbeber:
:+1:

aermakov-zalando:
👍

PR #347, opened 2021-07-24
fixes for ZMON client

pitr:
# One-line summary

Fixes to ZMON client post-July changes

## Description
- only add group_by "key" if key is provided
- remove deprecated aggregators "dev" and "first"
- add X-Attribution header for query attribution

## Types of Changes


mikkeloscar:
Could you also update the ZMON section in the README.md to not have the aggregators no longer available?

jonathanbeber:
thanks a lot for this @pitr, I created #348 just because I think we should at least minimally test it, since we weren't doing it before, I pushed some tests.

PR #349, opened 2021-07-31
Add support for scaling based on RouteGroup resource

mikkeloscar:
Extends the skipper collector to also support referencing a [`RouteGroup`](https://opensource.zalando.com/skipper/kubernetes/routegroups/) resource for scaling based on RPS.

The weights are based on the `defaultBackends` defined for the routegroup. It does not try to handle specific weighting per path or so which is possible in RouteGroup but is not the common use case. This would be compatible with the ingress RPS scaling we have on StackSets already.


pkg/collector/skipper_collector.go
```diff
@@ -128,26 +134,65 @@ func getWeights(ingressAnnotations map[string]string, backendAnnotations []strin
...
-	if err != nil {
-		return nil, err
+	for _, backend := range backends {
+		if backend.BackendName == backendName {
+			return float64(backend.Weight) / 100.0, nil
+		}
 	}
-	config := c.config
 
+	if backendName != "" {
```
jonathanbeber, line 148: Is it possible for `backend.BackendName` to be empty? I don't think so.
I'd suggest moving this check to the top and invert (i.e. `if backendName == ""`) and returning earlier, don't see why to range over all backends to figure out an empty backend name was provided.

pkg/collector/skipper_collector.go
```diff
@@ -128,26 +134,65 @@ func getWeights(ingressAnnotations map[string]string, backendAnnotations []strin
...
-		escapedHostnames = append(escapedHostnames, regexp.QuoteMeta(strings.Replace(rule.Host, ".", "_", -1)))
+	var backendWeight float64
+	switch c.objectReference.Kind {
+	case "Ingress":
+		ingress, err := c.client.NetworkingV1beta1().Ingresses(c.objectReference.Namespace).Get(ctx, c.objectReference.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+
+		backendWeight, err = getWeights(ingress.Annotations, c.backendAnnotations, c.backend)
```
jonathanbeber, line 166: :nail_care: nit: rename to 
```suggestion
		backendWeight, err = getIngressWeight(ingress.Annotations, c.backendAnnotations, c.backend)
```

pkg/collector/skipper_collector.go
mikkeloscar: Right, changed!

pkg/collector/skipper_collector.go
mikkeloscar: changed!

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #350, opened 2021-07-31
Use labels for specifying Ingress/RouteGroup backend

mikkeloscar:
When users wanted to scaled based on the RPS of a single backend for a traffic weighted Ingress/RouteGroup they could define the backend name as part of the metric name e.g. `requests-per-second,<backend>`. This "hack" of making the backend part of the metric name was introduced back when we used `autoscaling/v2beta1` when labels were not supported on Object type metrics.

Switch to use the match labels for defining the `backend` which is possible with `autoscaling/v2beta2`.

```yaml
metric:
  name: requests-per-second
  selector:
    matchLabels:
      backend: backend1 # optional backend
```

The previous deprecated format looked like this:

```yaml
metric:
  name: requests-per-second,backend1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,9 @@ spec:
         name: myapp
       metric:
         name: requests-per-second
+        selector:
+          matchLabels:
+            backend: backend1 # optional backend
       target:
         averageValue: "10"
         type: AverageValue
@@ -382,6 +385,9 @@ spec:
         name: myapp
       metric:
         name: requests-per-second
+        selector:
+          matchLabels:
+            backend: backend1 # optional backend
       target:
         averageValue: "10"
         type: AverageValue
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -382,22 +385,23 @@ spec:
...
-metric name like `requests-per-second,backend1` which would return the
-requests-per-second being sent to the `backend1`. The ingress annotation where
-the backend weights can be obtained can be specified through the flag
+Skipper supports sending traffic to different backends based on annotations
+present on the `Ingress` object, or weights on the RouteGroup backends. By
+default the number of replicas will be calculated based on the full traffic
+served by that ingress/routegroup.  If however only the traffic being routed to
+a specific backend should be used then the backend name can be specified via
+the `backend` label under `matchLabels` for the metric.  The ingress annotation
+where the backend weights can be obtained can be specified through the flag
```
jonathanbeber, line 404: Did you think about adding at least a deprecation message for the old format? I'm not sure if it's worth it, but at least to people understand when they come to see their config that they could/should migrate to the new format. At least I understand the old format is still valid for now.

pkg/collector/skipper_collector.go
```diff
@@ -51,11 +51,13 @@ func NewSkipperCollectorPlugin(client kubernetes.Interface, rgClient rginterface
...
 func (c *SkipperCollectorPlugin) NewCollector(hpa *autoscalingv2.HorizontalPodAutoscaler, config *MetricConfig, interval time.Duration) (Collector, error) {
 	if strings.HasPrefix(config.Metric.Name, rpsMetricName) {
-		backend := ""
-		if len(config.Metric.Name) > len(rpsMetricName) {
-			metricNameParts := strings.Split(config.Metric.Name, rpsMetricBackendSeparator)
-			if len(metricNameParts) == 2 {
-				backend = metricNameParts[1]
+		backend, ok := config.Config["backend"]
+		if !ok {
+			if len(config.Metric.Name) > len(rpsMetricName) {
```
jonathanbeber, line 56: What do you think about having a `TODO` to deprecate/cleanup it?

jonathanbeber:
two nitpicks. IMHO, the code looks good  and you have my +1 if you believe the topics I brought up are not relevant now.

README.md
mikkeloscar: I wouldn't put too much history in docs as it's usually confusing to newcomers. I would put it in the release notes pointing to this PR explaining what the change is.

pkg/collector/skipper_collector.go
mikkeloscar: done!

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #370, opened 2021-09-23
Scheduled scaling: scale up/down slowly

aermakov-zalando:
For the scheduled scaling, fade the configured value in and out slowly across a configured interval instead of jumping from 0 to the max immediately. This spreads the scale events around, creating less load on the other components, and helping the rest of the metrics (like the CPU ones) to adjust as well.

The duration of the scaling window can be configured individually in the CRD globally for all scheduled events, and defaults to a globally configured value if not specified. The default for the latter is set to 10 minutes, but can be changed using the `--scaling-schedule-default-scaling-window` flag.


aermakov-zalando:
👍

pkg/collector/scaling_schedule_collector_test.go
```diff
@@ -80,7 +86,45 @@ func TestScalingScheduleCollector(t *testing.T) {
 			expectedValue: 100,
 		},
 		{
-			msg: "Return the default value (0) for one time config - 30 seconds after",
+			msg: "Return the scaled value (50) for one time config - 20 seconds before starting",
```
mikkeloscar, line 89: nit: the comment doesn't align with the expected value: 50/67

mikkeloscar:
:+1:

pkg/collector/scaling_schedule_collector_test.go
aermakov-zalando: Oops, will fix. Thanks!

aermakov-zalando:
👍

mikkeloscar:
:+1:

PR #371, opened 2021-09-30
Use 10 buckets on ScalingSchedule ramp-up/down

jonathanbeber:
The HPA has a feature to do not scale up and down when the change in the
metric is less than 10%:

> We'll skip scaling if the ratio is sufficiently close to 1.0 (within a
> globally-configurable tolerance, from the
> `--horizontal-pod-autoscaler-tolerance` flag, which defaults to 0.1.

It could lead to pods scaling up to 10% less than the target for
ScalingSchedules and then not scaling to the actual value if the metric
calculated before was less than 10% of the target.

This commit uses 10 fixed buckets for scaling, this way we know the
metric returned during a scaling event is at least 10% more than a
previous one calculated during the period of ramp-up. The same is valid
for the scaling down during a ramp-down

### DIFF
--- a/pkg/collector/scaling_schedule_collector.go
+++ b/pkg/collector/scaling_schedule_collector.go
@@ -334,7 +334,14 @@ func scaledValue(timestamp time.Time, startTime time.Time, scalingWindowDuration
 	if scalingWindowDuration == 0 {
 		return 0
 	}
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

jonathanbeber:
:+1:

PR #376, opened 2021-11-23
Use labels hash in the custom metrics store and metrics store :nail_care: 

jonathanbeber:
#### Use labels hash in the custom metrics store 

The current implementation of the metrics store for custom metrics uses
just the object name as key. When scenarios where two HPAs reference the
same object but have different metric calculation (e.g. ingresses with
different weights), kube-metrics-adapter calculates the two metrics but
always overrides one of the metrics when saving it to the store.

This commit implements the use of a labels hash in the custom metrics
store. This way, metrics are identified not just by the referenced
object, but also by its selectors.

#### Define stronger types for metrics store 

The metrics store, both the custom and external one, make heavy usage of
maps of strings to strings. This leads to a presumable confusing or at
least hard to read codebase.

This commit tries to increase readability and maintainability of the
metric stores by using custom types. This commit has no effect in
performance or functionality, it's pure aesthetics.

### DIFF
--- a/pkg/provider/metric_store.go
+++ b/pkg/provider/metric_store.go
@@ -32,7 +32,7 @@ type externalMetricsStoredMetric struct {
 // MetricStore is a simple in-memory Metrics Store for HPA metrics.
 type MetricStore struct {
 	// metricName -> referencedResource -> objectNamespace -> objectName -> metric
-	customMetricsStore map[string]map[schema.GroupResource]map[string]map[string]customMetricsStoredMetric
+	customMetricsStore map[string]map[schema.GroupResource]map[string]map[string]map[string]customMetricsStoredMetric
 	// namespace -> metricName -> labels -> metric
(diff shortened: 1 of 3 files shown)


pkg/provider/metric_store.go
```diff
@@ -32,18 +32,33 @@ type externalMetricsStoredMetric struct {
 // MetricStore is a simple in-memory Metrics Store for HPA metrics.
 type MetricStore struct {
 	// metricName -> referencedResource -> objectNamespace -> objectName -> metric
```
mikkeloscar, line 34: I think this comment is wrong/less valuable now. It's missing the labelsHash and could maybe be dropped if the type changes make this part clear now.

mikkeloscar:
:+1:

jonathanbeber:
:+1:

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #380, opened 2022-01-12
Fix panic when trying to add key to nil map

katyanna:
**Fix panic when trying to add key to nil map**
Bug was introduced trying to solved the following issue:
> When scenarios where two HPAs reference the same object but have
different metric calculation (e.g. ingresses with different weights),
kube-metrics-adapter calculates the two metrics but always overrides
one of the metrics when saving it to the store. - https://github.com/zalando-incubator/kube-metrics-adapter/pull/376

This commit fixes a issue where without the added `return` the system
was continuing in an invalid states where `labels2metric` wasn't
properly initialized, causing the system to panic.

**Fix test filename to match related file**
Having the test filename prefix exactly as the tested file name helps systems
to identify the correlation.


mikkeloscar:
:+1:

katyanna:
👍

PR #398, opened 2022-02-07
Use cdp-runtime for build

mikkeloscar:
CDP is pushing a new concept called `runtime`s for setting up the build environment. It's similar to `overlay`s but is implemented slightly different (as a docker container as opposed to a tar extracting the content on the VM filesystem).

The runtime and overlay for go are pretty much identical in terms of tools installed etc.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -1,7 +1,9 @@
 version: "2017-09-20"
 pipeline:
 - id: build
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

Issue #400, opened 2022-02-14
Update outdated base image?


rgarcia89:
My container vulnerability scanner is showing multiple issues for the latest release, due to the fact, that it is based on `alpine-3.12`.
Are there any plans to update to alpine 3.15?

rgarcia89:
Awesome - thanks a lot for your fast reaction :)
Looks also good on my side now

PR #401, opened 2022-02-14
Update dependencies

mikkeloscar:
Update dependencies to address two CVEs and align with the automatic suggestions from @dependabot

Also update to the latest Zalando support alpine base image.


szuecs:
this should fix https://github.com/zalando-incubator/kube-metrics-adapter/issues/400

szuecs:
:+1:

PR #447, opened 2022-06-20
Limit ingress support to only networking.k8s.io/v1

katyanna:
Given the deprecation of `extensions/v1beta1`, remove it from Ingress possibilities.


katyanna:
:+1:

mikkeloscar:
With this, we will only support Kubernetes v1.19 and above for the skipper-collector. The README should reflect that: https://github.com/zalando-incubator/kube-metrics-adapter#skipper-collector

mikkeloscar:
:+1:

PR #505, opened 2022-12-13
Update scalingschedule config

zaklawrencea:
# Add end time/date options for (cluster) scaling schedules. 

## Description
The added end time/date options adds greater flexibility for (cluster) scaling schedules as this offers an alternative to the current start time/date + duration (in minutes) configuration - which is not easily readable when longer durations are present. The end time/date is an optional field which will only override the schedule duration if it is further in the future. Likewise, the duration field will override the scheduled end date if it is set to end after the scheduled end date. In short: the latest ending time/date will be given priority.

### DIFF
--- a/docs/cluster_scaling_schedules_crd.yaml
+++ b/docs/cluster_scaling_schedules_crd.yaml
@@ -55,8 +55,8 @@ spec:
                       format: date-time
                       type: string
                     durationMinutes:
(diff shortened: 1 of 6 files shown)


mikkeloscar:
:+1:

zaklawrencea:
👍

PR #506, opened 2022-12-21
Fix nil pointer dereference

zaklawrencea:
Fix nil pointer dereference when checking if an end date is set for a OneTime Schedule. Also add a test to check for an incorrect end date format.


gargravarr:
:+1:

demonCoder95:
👍

PR #538, opened 2023-03-01
Update dependencies

mikkeloscar:
Aggregated update of dependencies

* Also update base image to rely on new Alpine Linux version.


linki:
👍

PR #552, opened 2023-04-26
Add hostname RPS metric collector

lucastt:
# One-line summary

Adds a new metric collector that gets RPS metrics from a specific hostname.

## Description

The hostname RPS metrics will be useful to fetch custom metrics in a generic way, no metter which ingress is being used. It brings more generic functionality than Skipper and route group metric collector. The hostname RPS metric collector has prometheus collector as a dependency working as a wrapper around prometheus collector to reduce complexity when configuring HPA based on RPS metrics by hostname, this is why its necessary to enable prometheus collector in order to have hostname RPS enabled. Also, this new collector introduces a two new flags:

- `external-rps-metrics`: This flag is actually a feature flag, it controls weather this collector is enabled or not.
-  `external-rps-metric-name`: This flag specifies the right metric to collect data from, this is what makes this collector generic since a user can point to any metric associated with RPS by hostname.

## Types of Changes

- Configuration change

## Tasks

- [x] Create new hostname collector
- [x] Add support for N hostnames per metric
- [x] Add support for hostname traffic weight, the same as skipper collector
- [x] Change metric name to `requests-per-second`
- [x] Update documentation
- [x] Test changes in pet cluster

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes

To activate this functionality is necessary to use the flag `hostname-rps-metrics` and have prometheus metric collector already enabled. If the cluster uses a ingress other than `skipper-ingress`, it is necessary to specify from which prometheus metric the collector should fetch the metrics using the flag `hostname-rps-metric-name`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,7 @@ metadata:
   name: myapp-hpa
   annotations:
     metric-config.external.example-rps.requests-per-second/hostname: www.example1.com,www.example2.com
-    metric-config.external.example-rps.requests-per-second/weight: 42
+    metric-config.external.example-rps.requests-per-second/weight: "42"
 spec:
   scaleTargetRef:
     apiVersion: apps/v1
--- a/README.md
+++ b/README.md
@@ -402,12 +402,12 @@ the `backend` label under `matchLabels` for the metric.  The ingress annotation
 where the backend weights can be obtained can be specified through the flag
 `--skipper-backends-annotation`.
 
-## Hostname RPS collector
+## External RPS collector
 
-The Hostname collector, like Skipper collector, is a simple wrapper around the Prometheus collector to
+The External RPS collector, like Skipper collector, is a simple wrapper around the Prometheus collector to
 make it easy to define an HPA for scaling based on the RPS measured for a given hostname. When
 [skipper](https://github.com/zalando/skipper) is used as the ingress
-implementation in your cluster everything should work automatically, in case another reverse proxy is used as ingress, like [Nginx](https://github.com/kubernetes/ingress-nginx) for example, its necessary to configure which prometheus metric should be used through `--hostname-rps-metric-name <metric-name>` flag. Assuming `skipper-ingress` is being used or the appropriate metric name is passed using the flag mentioned previously this collector provides the correct Prometheus queries out of the
+implementation in your cluster everything should work automatically, in case another reverse proxy is used as ingress, like [Nginx](https://github.com/kubernetes/ingress-nginx) for example, its necessary to configure which prometheus metric should be used through `--external-rps-metric-name <metric-name>` flag. Assuming `skipper-ingress` is being used or the appropriate metric name is passed using the flag mentioned previously this collector provides the correct Prometheus queries out of the
 box so users don't have to define those manually.
 
 ### Supported metrics
@@ -421,7 +421,7 @@ box so users don't have to define those manually.
 This is an example of an HPA that will scale based on `requests-per-second` for the RPS measured in the hostnames called: `www.example1.com` and `www.example2.com`; and weighted by 42%.
 
 ```yaml
-apiVersion: autoscaling/v2beta2
+apiVersion: autoscaling/v2
 kind: HorizontalPodAutoscaler
 metadata:
   name: myapp-hpa
@@ -449,7 +449,7 @@ spec:
 ```
 ### Multiple hostnames per metric
 
-This metric supports a relation of n:1 with metrics. The way it works is the measured RPS is the sum of the RPS rate of each of the specified hostnames. This value is further modified by the weight parameter explained bellow.
+This metric supports a relation of n:1 between hostnames and metrics. The way it works is the measured RPS is the sum of the RPS rate of each of the specified hostnames. This value is further modified by the weight parameter explained bellow.
 
 ### Metric weighting based on backend
 
--- a/README.md
+++ b/README.md
@@ -414,11 +414,11 @@ box so users don't have to define those manually.
 
 | Metric | Description | Type | Kind | K8s Versions |
 | ------------ | -------------- | ------- | -- | -- |
-| `hostname-rps` | Scale based on requests per second for a certain hostname. | External | | `>=1.12` |
+| `requests-per-second` | Scale based on requests per second for a certain hostname. | External | | `>=1.12` |
 
 ### Example: External Metric
 
-This is an example of an HPA that will scale based on `hostname-rps` for the RPS measured in the hostname called `www.example.com`.
+This is an example of an HPA that will scale based on `requests-per-second` for the RPS measured in the hostnames called: `www.example1.com` and `www.example2.com`; and weighted by 42%.
 
 ```yaml
 apiVersion: autoscaling/v2beta2
@@ -426,7 +426,8 @@ kind: HorizontalPodAutoscaler
 metadata:
   name: myapp-hpa
   annotations:
-    metric-config.external.example-rps.hostname-rps/hostname: www.example.com
+    metric-config.external.example-rps.requests-per-second/hostname: www.example1.com,www.example2.com
+    metric-config.external.example-rps.requests-per-second/weight: 42
 spec:
   scaleTargetRef:
     apiVersion: apps/v1
@@ -441,11 +442,23 @@ spec:
         name: example-rps
         selector:
           matchLabels:
-            type: hostname-rps
+            type: requests-per-second
       target:
         type: AverageValue
         averageValue: "42"
 ```
+### Multiple hostnames per metric
+
+This metric supports a relation of n:1 with metrics. The way it works is the measured RPS is the sum of the RPS rate of each of the specified hostnames. This value is further modified by the weight parameter explained bellow.
+
+### Metric weighting based on backend
+
+There are ingress-controllers, like skipper-ingress, that supports sending traffic to different backends based on some kind of configuration, in case of skipper annotations
+present on the `Ingress` object, or weights on the RouteGroup backends. By
+default the number of replicas will be calculated based on the full traffic
+served by these components.  If however only the traffic being routed to
+a specific hostname should be used then the weight for the configured hostname(s) might be specified via the `weight` annotation `metric-config.external.<metric-name>.request-per-second/weight` for the metric being configured.
+
 
 ## InfluxDB collector
 
--- a/README.md
+++ b/README.md
@@ -402,6 +402,51 @@ the `backend` label under `matchLabels` for the metric.  The ingress annotation
 where the backend weights can be obtained can be specified through the flag
 `--skipper-backends-annotation`.
 
+## Hostname RPS collector
+
+The Hostname collector, like Skipper collector, is a simple wrapper around the Prometheus collector to
+make it easy to define an HPA for scaling based on the RPS measured for a given hostname. When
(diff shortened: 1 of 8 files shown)


pkg/server/start.go
```diff
@@ -65,6 +65,7 @@ func NewCommandStartAdapterServer(stopCh <-chan struct{}) *cobra.Command {
 		MetricsAddress:                    ":7979",
 		ZMONTokenName:                     "zmon",
 		CredentialsDir:                    "/meta/credentials",
+		HostnameRPSMetricName:             "skipper_serve_host_duration_seconds_count",
```
lucastt, line 68: This might be a mistake, if I only activate this collector when I pass the metric name, I should not have a default metric name. Something I could do to have a default metric name is to have a second flag that the only purpose is to activate this collector. The other option would be not having a default value for this flag. I'll decide it as the PR progresses.

pkg/collector/hostname_collector.go
```diff
@@ -0,0 +1,93 @@
...
+	v, err := c.promCollector.GetMetrics()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(v) != 1 {
+		return nil, fmt.Errorf("expected to only get one metric value, got %d", len(v))
+	}
+
+	// TBD(Lucas):
```
lucastt, line 80: This will not be necessary, talking with @mikkeloscar object metrics don't need to be supported here only external metrics.

README.md
```diff
@@ -404,6 +404,51 @@ the `backend` label under `matchLabels` for the metric.  The ingress annotation
 where the backend weights can be obtained can be specified through the flag
 `--skipper-backends-annotation`.
 
+## Hostname RPS collector
```
mikkeloscar, line 407: Considering that this feature is supposed to cover what the existing Object based `request-per-second` feature is doing I would consider naming it just `request-per-second` (more generic) and allow multiple hostnames to be defined. This way it would cover a case we have in Zalando where a single application is served by multiple hostnames e.g. `app.example.com` and `app.ingress.cluster.local` (internal to the cluster). In the Object based metric we aggregate the RPS over all hostnames (see skipper collector) we could do the same here giving the users the option to define one or more hostnames.

Another feature we provide which is relevant in combination with StackSet is scaling the RPS based on traffic weight. E.g. if an application has 1000 RPS and it has two versions where one is getting 30% and the other 70% then we scale the RPS like `1000 * 30%` and `1000 * 70%` respectively, such that the stacks don't over scale during the traffic switching.
We could acheive the same if we allow passing a `scale` config (or a better name?) which is used as part of the query. The value of the `scale` config can be set by StackSet controller or manually by users if they don't use StackSet, but that use-case is less likely to be relevant.

lucastt: Ok, so the idea would be to do the same as skipper collector do, but the metric should be of type `external` and be hostname based, instead of `ingress`/`routegroup` based. Since the type - skipper is `object` and hostname is `external` - is different it should be no problem to have two collectors with the same name. Did I understood everything correctly?

mikkeloscar: > is different it should be no problem to have two collectors with the same name. Did I understood everything correctly?

Yes, it should be no problem using the same name.

lucastt: Sounds good.

pkg/collector/hostname_collector.go
```diff
@@ -0,0 +1,116 @@
...
+	interval time.Duration,
+) (Collector, error) {
+	if config == nil {
+		return nil, fmt.Errorf("Metric config not present, it is not possible to initialize the collector.")
+	}
+	// Need to copy config and add a promQL query in order to get
+	// RPS data from a specific hostname from prometheus. The idea
+	// of the copy is to not modify the original config struct.
+	confCopy := *config
+	//hostnames := config.Config["hostnames"]
```
lucastt, line 55: TODO (Lucas): Need to clean this up

README.md
mikkeloscar: Let's change the title here to be just `RPS Collector`?

pkg/collector/hostname_collector.go
```diff
@@ -0,0 +1,119 @@
+package collector
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	//autoscalingv2 "k8s.io/api/autoscaling/v2beta2"
```
mikkeloscar, line 10: nit: remove this commented line :)

pkg/collector/hostname_collector.go
```diff
@@ -0,0 +1,119 @@
...
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+)
+
+const (
+	HostnameMetricType = "requests-per-second"
+	HostnameRPSQuery   = `scalar(sum(rate(%s{host=~"%s"}[1m])) * %.4f)`
+)
+
+type HostnameCollectorPlugin struct {
```
mikkeloscar, line 19: Let's change `Hostname` to `RPS` in all these cases. Same with the filename `rps_collector.go`?

pkg/collector/hostname_collector.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	// Need to copy config and add a promQL query in order to get
+	// RPS data from a specific hostname from prometheus. The idea
+	// of the copy is to not modify the original config struct.
+	confCopy := *config
+
+	if _, ok := config.Config["hostnames"]; !ok {
+		return nil, fmt.Errorf("Hostname is not specified, unable to create collector")
+	}
+	regex, err := regexp.Compile("^[a-zA-Z0-9.-]+$")
```
mikkeloscar, line 60: nit: I think we have compile this on as a package global variable and avoid compiling it for every initialization.

pkg/collector/hostname_collector.go
lucastt: Oh, sorry, I missed this one

pkg/collector/hostname_collector.go
lucastt: I'm ok with it, just worried to people confusing it with skipper_collector RPS metric in the future. I know it is different, but for someone relatively new to the code it might be confusing, if you think this is ok I don't mind changing it.

pkg/collector/hostname_collector.go
lucastt: Oh, cool, I'll check it

pkg/collector/hostname_collector.go
mikkeloscar: It's not a big deal to me. If you think it's easier to understand as it is, then that is ok from my perspective.

I know the code base well so I may be biased 🙃

pkg/collector/hostname_collector.go
mikkeloscar: I don't see what you describe reflected in the code. Are you just proposing it?

I think the change of it being misused if it's a package local variable is very low. It's essentially used as a constant, but can't be constant because of the type.

Initializing it in the factory seems ok as it would only initialize it once, but you still need to pass it on plugin initialization which is maybe a bit overkill considering the _constant_ nature of the variable. My main point was to not compile the regex multiple times for no reason as it's not dynamic in any way.

pkg/collector/hostname_collector.go
lucastt: I think your idea is better, but I'm still worried with name confusion, what do you think about `HostnameRPSCollector` or `ExternalRPSCollector`, the second emphasising the fact this is the collector for the `External` type?

mikkeloscar: `ExternalRPSCollector` I like that

lucastt:
👍

lucastt:
Tested in pet-cluster using following config:

hpa.yaml:
```yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: skipper-app-hpa
  namespace: default
  annotations:
    metric-config.external.skipper-app-rps.requests-per-second/hostnames: <test-hostname>
    metric-config.external.skipper-app-rps.requests-per-second/weight: "100"
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: skipper-app
  minReplicas: 1
  maxReplicas: 10
  metrics:
  - type: External
    external:
      metric:
        name: skipper-app-rps
        selector:
          matchLabels:
            type: requests-per-second
      target:
        type: AverageValue
        averageValue: "2"
```
generated load using [k6](https://k6.io/docs/#what-is-k6) with the following script:

```js
import http from 'k6/http';
import { check, sleep } from 'k6';

export const options = {
  vus: 20,
  duration: '10m',
};

export default function () {
  const res = http.get('<test-hostname>');
  check(res, { 'status was 200': (r) => r.status == 200 });
  sleep(1);
}
```

and got the following result in the cluster (it started as just one replica):

```
$ k get pods
NAME                           READY   STATUS    RESTARTS   AGE
skipper-app-86589444b7-22hs4   1/1     Running   0          68s
skipper-app-86589444b7-4sc2n   1/1     Running   0          38s
skipper-app-86589444b7-hcrwt   1/1     Running   0          8s
skipper-app-86589444b7-kzq4d   1/1     Running   0          39m
skipper-app-86589444b7-lh87q   1/1     Running   0          68s
skipper-app-86589444b7-pg8jd   1/1     Running   0          38s
skipper-app-86589444b7-tztj9   1/1     Running   0          68s
skipper-app-86589444b7-vq9rs   1/1     Running   0          38s
skipper-app-86589444b7-ww2wk   1/1     Running   0          38s
skipper-app-86589444b7-zf252   1/1     Running   0          8s
```

when the script finished running:

```
$ k get pods   
NAME                           READY   STATUS    RESTARTS   AGE
skipper-app-86589444b7-vq9rs   1/1     Running   0          9m53s
```

Everything looks like working as expected.

lucastt:
👍

README.md
```diff
@@ -402,6 +402,64 @@ the `backend` label under `matchLabels` for the metric.  The ingress annotation
...
+This is an example of an HPA that will scale based on `requests-per-second` for the RPS measured in the hostnames called: `www.example1.com` and `www.example2.com`; and weighted by 42%.
+
+```yaml
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: myapp-hpa
+  annotations:
+    metric-config.external.example-rps.requests-per-second/hostname: www.example1.com,www.example2.com
+    metric-config.external.example-rps.requests-per-second/weight: 42
```
mikkeloscar, line 430:
```suggestion
    metric-config.external.example-rps.requests-per-second/weight: "42"
```

mikkeloscar:
:+1:

lucastt:
👍

README.md
```diff
@@ -402,6 +402,64 @@ the `backend` label under `matchLabels` for the metric.  The ingress annotation
...
+
+This is an example of an HPA that will scale based on `requests-per-second` for the RPS measured in the hostnames called: `www.example1.com` and `www.example2.com`; and weighted by 42%.
+
+```yaml
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: myapp-hpa
+  annotations:
+    metric-config.external.example-rps.requests-per-second/hostname: www.example1.com,www.example2.com
```
AlexanderYastrebov, line 429: Looks like this must be plural `hostnames`

PR #636, opened 2023-11-15
Add support for scaling based on Nakadi

### LINKED ISSUE #5: idea: Nakadi collector
Reference: https://nakadi.io/manual.html#/subscriptions/subscription_id/stats_get

> We have a service that pushes unconsumed events into CloudWatch (in GHE: tea/nakadiwatchdog), this may help to get started.

https://github.com/mikkeloscar/kube-metrics-adapter/issues/1#issuecomment-404738138

mikkeloscar:
This adds support for scaling based on Nakadi subscriptions by utilizing the nakadi stats api: https://nakadi.io/manual.html#/subscriptions/subscription_id/stats_get

It's possible to define an HPA like below:

```yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: myapp-hpa
  annotations:
    metric-config.external.my-nakadi-consumer.nakadi/interval: "60s" # optional
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: custom-metrics-consumer
  minReplicas: 0
  maxReplicas: 8 # should match number of partitions for the event type
  metrics:
  - type: External
    external:
      metric:
          name: my-nakadi-consumer
          selector:
            matchLabels:
              type: nakadi
              subscription-id: "708095f6-cece-4d02-840e-ee488d710b29"
              metric-type: "consumer-lag-seconds|unconsumed-events"
      target:
        # value is compatible with the consumer-lag-seconds metric type.
        # It describes the amount of consumer lag in seconds before scaling
        # additionally up.
        # if an event-type has multiple partitions the value of
        # consumer-lag-seconds is the max of all the partitions.
        value: "600" # 10m
        # averageValue is compatible with unconsumed-events metric type.
        # This means for every 30 unconsumed events a pod is scaled up.
        # unconsumed-events is the sum of of unconsumed_events over all
        # partitions.
        averageValue: "30"
        type: AverageValue
```

To scale on either `consumer-lag-seconds` or `unconsumed-events`.

Fix #5

## TODO

* [x] Unit tests
* [x] Better documentation in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -706,14 +706,17 @@ defined, then the annotation takes precedence.
 ## Nakadi collector
 
 The Nakadi collector allows scaling based on [Nakadi](https://nakadi.io/)
-`consumer_lag_seconds` or `unconsumed_events`.
+ Subscription API stats metrics `consumer_lag_seconds` or `unconsumed_events`.
 
 ### Supported metrics
 
-| Metric Type | Description | Type | K8s Versions |
+| Metric Type            | Description                                                                 | Type     | K8s Versions |
+|------------------------|-----------------------------------------------------------------------------|----------|--------------|
+| `unconsumed-events`    | Scale based on number of unconsumed events for a Nakadi subscription        | External | `>=1.24`     |
+| `consumer-lag-seconds` | Scale based on number of max consumer lag seconds for a Nakadi subscription | External | `>=1.24`     |
 | ------------ | ------- | -- | -- |
-| `unconsumed-events` | Scale based on number of unconsumed events for a Nakadi subscription | External | `>=1.24` |
-| `consumer-lag-seconds` | Scale based on number of max consumer lag seconds for a Nakadi subscription | External | `>=1.24` |
+| `unconsumed-events` | Scale based on number of unconsumed events for a Nakadi Subscription | External | `>=1.24` |
+| `consumer-lag-seconds` | Scale based on number of max consumer lag seconds for a Nakadi Subscription | External | `>=1.24` |
 
 ```yaml
 apiVersion: autoscaling/v2
@@ -748,14 +751,14 @@ spec:
         # consumer-lag-seconds is the max of all the partitions.
         value: "600" # 10m
         # averageValue is compatible with unconsumed-events metric type.
-        # This means for every 30 unconsumed events a pod is scaled up.
+        # This means for every 30 unconsumed events a pod is added.
         # unconsumed-events is the sum of of unconsumed_events over all
         # partitions.
         averageValue: "30"
         type: AverageValue
 ```
 
-The `subscription-id` is the ID of the relevant consumer subscription. The
+The `subscription-id` is the Subscription ID of the relevant consumer. The
 `metric-type` indicates whether to scale on `consumer-lag-seconds` or
 `unconsumed-events` as outlined below.
 
--- a/README.md
+++ b/README.md
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
 metric label definitions. If both annotations and corresponding label is
 defined, then the annotation takes precedence.
 
+
+## Nakadi collector
+
+The Nakadi collector allows scaling based on [Nakadi](https://nakadi.io/)
+`consumer_lag_seconds` or `unconsumed_events`.
+
+### Supported metrics
+
+| Metric Type | Description | Type | K8s Versions |
+| ------------ | ------- | -- | -- |
+| `unconsumed-events` | Scale based on number of unconsumed events for a Nakadi subscription | External | `>=1.24` |
+| `consumer-lag-seconds` | Scale based on number of max consumer lag seconds for a Nakadi subscription | External | `>=1.24` |
+
+```yaml
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: myapp-hpa
+  annotations:
+    # metric-config.<metricType>.<metricName>.<collectorType>/<configKey>
+    metric-config.external.my-nakadi-consumer.nakadi/interval: "60s" # optional
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: custom-metrics-consumer
+  minReplicas: 0
+  maxReplicas: 8 # should match number of partitions for the event type
+  metrics:
+  - type: External
+    external:
+      metric:
+          name: my-nakadi-consumer
+          selector:
+            matchLabels:
+              type: nakadi
+              subscription-id: "708095f6-cece-4d02-840e-ee488d710b29"
+              metric-type: "consumer-lag-seconds|unconsumed-events"
+      target:
+        # value is compatible with the consumer-lag-seconds metric type.
+        # It describes the amount of consumer lag in seconds before scaling
+        # additionally up.
+        # if an event-type has multiple partitions the value of
+        # consumer-lag-seconds is the max of all the partitions.
+        value: "600" # 10m
+        # averageValue is compatible with unconsumed-events metric type.
+        # This means for every 30 unconsumed events a pod is scaled up.
+        # unconsumed-events is the sum of of unconsumed_events over all
+        # partitions.
+        averageValue: "30"
+        type: AverageValue
+```
+
+The `subscription-id` is the ID of the relevant consumer subscription. The
+`metric-type` indicates whether to scale on `consumer-lag-seconds` or
+`unconsumed-events` as outlined below.
+
+`unconsumed-events` - is the total number of unconsumed events over all
+partitions. When using this `metric-type` you should also use the target
+`averageValue` which indicates the number of events which can be handled per
+pod. To best estimate the number of events per pods, you need to understand the
+average time for processing an event as well as the rate of events.
+
+*Example*: You have an event type producing 100 events per second between 00:00
+and 08:00. Between 08:01 to 23:59 it produces 400 events per second.
+Let's assume that on average a single pod can consume 100 events per second,
+then we can define 100 as `averageValue` and the HPA would scale to 1 between
+00:00 and 08:00, and scale to 4 between 08:01 and 23:59. If there for some
+reason is a short spike of 800 events per second, then it would scale to 8 pods
+to process those events until the rate goes down again.
+
+`consumer-lag-seconds` -  describes the age of the oldest unconsumed event for
+a subscription. If the event type has multiple partitions the lag is defined as
+the max age over all partitions. When using this `metric-type` you should use
+the target `value` to indicate the max lag (in seconds) before the HPA should
+scale.
+
+*Example*: You have a subscription with a defined SLO of "99.99 of events are
+consumed within 30 min.". In this case you can define a target `value` of e.g.
+20 min. (1200s) (to include a safety buffer) such that the HPA only scales up
+from 1 to 2 if the target of 20 min. is breached and it needs to work faster
+with more consumers.
+For this case you should also account for the average time for processing an
+event when defining the target.
+
+
 ## HTTP Collector
 
 The http collector allows collecting metrics from an external endpoint specified in the HPA.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -702,6 +702,51 @@ you need to define a `key` or other `tag` with a "star" query syntax like
...
+  maxReplicas: 8 # should match number of partitions for the event type
+  metrics:
+  - type: External
+    external:
+      metric:
+          name: nakadi-consumer
+          selector:
+            matchLabels:
+              type: nakadi
+              subscription-id: "708095f6-cece-4d02-840e-ee488d710b29"
```
AlexanderYastrebov, line 733: I think subscription is uniquely identified by the `owning_application`, `event_types` and `consumer_group` https://nakadi.io/manual.html#creating-subscriptions so maybe this metric handler could also support these three fields and figure out subscription id by listing subscriptions https://nakadi.io/manual.html#getting-and-listing-subscriptions

This could be handy to define the same logical subscription (same `owning_application`, `event_types` and `consumer_group`) in staging and production environments

pkg/nakadi/nakadi.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return unconsumedEvents, nil
+}
+
+type statsResp struct {
+	Items []statsSubscription `json:"items"`
+}
+
+type statsSubscription struct {
```
AlexanderYastrebov, line 66: I think this should be statsEventType https://nakadi.io/manual.html#definition_SubscriptionEventTypeStats

pkg/nakadi/nakadi.go
```diff
@@ -0,0 +1,125 @@
...
+	err = json.Unmarshal(d, &result)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result.Items) != 1 {
+		return nil, fmt.Errorf("expected 1 subscription, %d returned", len(result.Items))
+	}
+
+	return &result.Items[0], nil
```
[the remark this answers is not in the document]
AlexanderYastrebov: There is a check above so it looks like only subscriptions for single event type are supported. Maybe this should be explicitly documented.

pkg/nakadi/nakadi.go
AlexanderYastrebov: Maybe add unused `event_type` field for clarity.

pkg/nakadi/nakadi.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	var result statsResp
+	err = json.Unmarshal(d, &result)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result.Items) != 1 {
+		return nil, fmt.Errorf("expected 1 subscription, %d returned", len(result.Items))
```
AlexanderYastrebov, line 121: expected 1 event type

pkg/nakadi/nakadi.go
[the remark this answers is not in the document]
decmurphy: Should be easy enough to extend this to multiple event-type subscriptions by specifying an `event-type` under `matchLabels`. If no `event-type` is specified then it could default to first in list, but this behaviour would need to be documented either way 👍

pkg/nakadi/nakadi.go
mikkeloscar: Changed!

pkg/nakadi/nakadi.go
mikkeloscar: updated

pkg/nakadi/nakadi.go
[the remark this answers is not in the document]
mikkeloscar: Extended to multiple event types.

It will aggregate over all event types for a subscription by default. Users can optionally add `event-type` to filter for a specific one.

pkg/nakadi/nakadi.go
```diff
@@ -0,0 +1,144 @@
...
+	var result statsResp
+	err = json.Unmarshal(d, &result)
+	if err != nil {
+		return nil, err
+	}
+
+	var eventTypes []statsEventType
+	if eventType != "" {
+		for _, et := range result.Items {
+			et := et
```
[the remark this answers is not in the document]
mikkeloscar: removed this logic again as we don't see a real use case to filter by event-type.

pkg/collector/nakadi_collector.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	metricValue := CollectedMetric{
+		Namespace: c.namespace,
+		Type:      c.metricType,
+		External: external_metrics.ExternalMetricValue{
+			MetricName:   c.metric.Name,
+			MetricLabels: c.metric.Selector.MatchLabels,
+			Timestamp:    metav1.Time{Time: time.Now().UTC()},
+			Value:        *resource.NewMilliQuantity(int64(value*1000), resource.DecimalSI),
```
AlexanderYastrebov, line 110: Why not NewQuantity here?
Also why not return int64 from c.nakadi to avoid type conversion?

mikkeloscar: Changed both.

pkg/collector/nakadi_collector.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+
+	metricValue := CollectedMetric{
+		Namespace: c.namespace,
+		Type:      c.metricType,
+		External: external_metrics.ExternalMetricValue{
+			MetricName:   c.metric.Name,
+			MetricLabels: c.metric.Selector.MatchLabels,
+			Timestamp:    metav1.Time{Time: time.Now().UTC()},
```
AlexanderYastrebov, line 109: Why not `metav1.Now()`? I think this is used eveywhere in k8s codebase

pkg/nakadi/nakadi.go
```diff
@@ -0,0 +1,131 @@
...
+func (c *Client) ConsumerLagSeconds(ctx context.Context, subscriptionID string) (int64, error) {
+	stats, err := c.stats(ctx, subscriptionID)
+	if err != nil {
+		return 0, err
+	}
+
+	var maxConsumerLagSeconds int64
+	for _, eventType := range stats {
+		for _, partition := range eventType.Partitions {
+			if partition.ConsumerLagSeconds > maxConsumerLagSeconds {
```
AlexanderYastrebov, line 42: minor: there is a builtin https://pkg.go.dev/builtin#max

pkg/nakadi/nakadi.go
```diff
@@ -0,0 +1,131 @@
...
+	q := endpoint.Query()
+	q.Set("show_time_lag", "true")
+	endpoint.RawQuery = q.Encode()
+
+	req, err := http.NewRequest(http.MethodGet, endpoint.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.http.Do(req)
```
AlexanderYastrebov, line 105: minor: c.http.Get(endpoint.String())?

pkg/collector/nakadi_collector.go
mikkeloscar: Fixed

pkg/nakadi/nakadi.go
mikkeloscar: Changed! I believe this is how you meant right?

pkg/nakadi/nakadi.go
mikkeloscar: Thanks! Changed!

AlexanderYastrebov:
:+1:

README.md
```diff
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
 metric label definitions. If both annotations and corresponding label is
 defined, then the annotation takes precedence.
 
+
+## Nakadi collector
+
+The Nakadi collector allows scaling based on [Nakadi](https://nakadi.io/)
+`consumer_lag_seconds` or `unconsumed_events`.
```
adyach, line 709:
```suggestion
 Subscription API stats metric as `consumer_lag_seconds` or `unconsumed_events`.
```

README.md
```diff
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
...
+
+### Supported metrics
+
+| Metric Type | Description | Type | K8s Versions |
```
adyach, line 713: | Metric Type            | Description                                                                 | Type     | K8s Versions |
|------------------------|-----------------------------------------------------------------------------|----------|--------------|
| `unconsumed-events`    | Scale based on number of unconsumed events for a Nakadi subscription        | External | `>=1.24`     |
| `consumer-lag-seconds` | Scale based on number of max consumer lag seconds for a Nakadi subscription | External | `>=1.24`     |


aligned table :)

adyach, line 713:
```suggestion
| Metric Type            | Description                                                                 | Type     | K8s Versions |
|------------------------|-----------------------------------------------------------------------------|----------|--------------|
| `unconsumed-events`    | Scale based on number of unconsumed events for a Nakadi subscription        | External | `>=1.24`     |
| `consumer-lag-seconds` | Scale based on number of max consumer lag seconds for a Nakadi subscription | External | `>=1.24`     |
```

adyach: aligned table looks always better :)

README.md
```diff
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
...
+| ------------ | ------- | -- | -- |
+| `unconsumed-events` | Scale based on number of unconsumed events for a Nakadi subscription | External | `>=1.24` |
```
adyach, line 715:
```suggestion
| `unconsumed-events` | Scale based on number of unconsumed events for a Nakadi Subscription | External | `>=1.24` |
```

README.md
```diff
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
...
+  annotations:
+    # metric-config.<metricType>.<metricName>.<collectorType>/<configKey>
+    metric-config.external.my-nakadi-consumer.nakadi/interval: "60s" # optional
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: custom-metrics-consumer
+  minReplicas: 0
+  maxReplicas: 8 # should match number of partitions for the event type
```
adyach, line 732: maybe we can improve it in the future that autoscaler does that on its own. it is possible to know number of partitions for Subscription

README.md
```diff
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
...
+              metric-type: "consumer-lag-seconds|unconsumed-events"
+      target:
+        # value is compatible with the consumer-lag-seconds metric type.
+        # It describes the amount of consumer lag in seconds before scaling
+        # additionally up.
+        # if an event-type has multiple partitions the value of
+        # consumer-lag-seconds is the max of all the partitions.
+        value: "600" # 10m
+        # averageValue is compatible with unconsumed-events metric type.
+        # This means for every 30 unconsumed events a pod is scaled up.
```
adyach, line 751:
```suggestion
        # This means for every 30 unconsumed events an additional pod is added.
```

README.md
```diff
@@ -702,6 +702,92 @@ you need to define a `key` or other `tag` with a "star" query syntax like
...
+        type: AverageValue
+```
+
+The `subscription-id` is the ID of the relevant consumer subscription. The
+`metric-type` indicates whether to scale on `consumer-lag-seconds` or
+`unconsumed-events` as outlined below.
+
+`unconsumed-events` - is the total number of unconsumed events over all
+partitions. When using this `metric-type` you should also use the target
+`averageValue` which indicates the number of events which can be handled per
```
adyach, line 764: is that really per pod or per the whole the stack? so if I have 2 pods and average number goes beyond 30 then a pod is added. it means it is 30 per 2 pods.

mikkeloscar: I guess this is how you define it. The autoscaler will always scale higher rather than lower. So if the averageTarget is 30 and you have 31 unconsumed events, it will scale to 2 pods, but it will not scale higher until 61 unconsumed events where it then needs 3 pods.

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #669, opened 2024-02-01
Update deps

mikkeloscar:
Aggregated update of deps

Most significant upgrading to Kubernetes v1.25 which is soon the default in Zalando clusters.


mikkeloscar:
:+1:

katyanna:
👍

PR #713, opened 2024-05-21
Introduce context for collector interface

mikkeloscar:
Many of the collectors are internally calling functions/methods which expect a context. In those cases we usually pass `context.TODO()` because there is no context passed from the caller.

Introduce context to the methods `NewCollector` and `GetMetrics` such that it's easier to use a passed context. For now we still pass `context.TODO()` from the caller, but eventually we can use a context with timeout where it makes sense.


mikkeloscar:
:+1:

katyanna:
👍

PR #729, opened 2024-06-18
Group go.mod updates

mikkeloscar:
Try out grouping dependabot updates

Disabled the schedule for now to test how it work. Ignoring `k8s.io/` packages as we normally update those manually together with kubernetes-on-aws.

ref: https://github.com/dependabot/dependabot-core/issues/7547


linki:
👍

PR #739, opened 2024-07-04
Only let Dependabot update weekly

mikkeloscar:
daily updates are a bit too spammy, only allow weekly updates.


mikkeloscar:
:+1:

katyanna:
👍

PR #765, opened 2024-12-02
Adjust HPA when schedule scaling change is below tolerance

mikkeloscar:
The Horizontal Pod Autoscaler (in `kube-controller-manager`) has a configuration `--horizontal-pod-autoscaler-tolerance=0.1` which defines how much a metric has to change before the HPA scales. E.g. if a metric is indicating a change that would result in less than 10% (default) change of replicas it won't take any action. This setting is meaningful for e.g. CPU or Memory based scaling where too frequent up and down-scale is not desired.

However, for a more stable metric, like Schedule Scaling, it's undesirable/confusing when it doesn't adjust the scaling when the change is below the tolerance.

Example: Imagine you have configured scheduled scaling to scale to 95 pods. Now the schedule is increased such that it will result in 100 pods. Because this is a 5.3% increase which is less than the default 10% it won't result in scaling and it will stay at 95 pods which is confusing to users.

This PR extends the schedule-scaling controller to go through all HPAs. For each it will check for active schedule scaling and the expected scale. If the expected scale is higher than the current HPA scale and the increase/change between current and expected is less than `--horizontal-pod-autoscaler-tolerance` then it will scale up to the expected scale.

Note: This is only implemented in one direction, scaling up, scaling down is significantly harder to implement in a correct/safe way because scaling down depends on knowning the desired scale of all other metrics of the HPA before making a decision if it's safe to scale down. Scaling up is simple because we can assume that the current HPA scale is the max of all metrics, otherwise it would have scaled higher already.

### DIFF
--- a/pkg/controller/scheduledscaling/scaler.go
+++ b/pkg/controller/scheduledscaling/scaler.go
@@ -1,0 +1,118 @@
+package scheduledscaling
+
+import (
+	"fmt"
+	"time"
+
+	"golang.org/x/net/context"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/client-go/scale"
+	scaleclient "k8s.io/client-go/scale"
+)
+
+// TargetScaler is an interface for scaling a target referenced resource in an
+// HPA to the desired replicas.
+type TargetScaler interface {
+	Scale(ctx context.Context, hpa *autoscalingv2.HorizontalPodAutoscaler, replicas int32) error
+}
+
+type hpaTargetScaler struct {
+	mapper      apimeta.RESTMapper
+	scaleClient scaleclient.ScalesGetter
+}
+
+// NewHPATargetScaler creates a new TargetScaler that can scale resources
+// targeted by HPAs. It takes a Kubernetes client and a REST config and uses a
+// restmapper to resolve the target reference API.
+func NewHPATargetScaler(ctx context.Context, kubeClient kubernetes.Interface, cfg *rest.Config) (TargetScaler, error) {
+	cachedClient := cacheddiscovery.NewMemCacheClient(kubeClient.Discovery())
+	restMapper := restmapper.NewDeferredDiscoveryRESTMapper(cachedClient)
+	go wait.Until(func() {
+		restMapper.Reset()
+	}, 30*time.Second, ctx.Done())
+
+	scaleKindResolver := scale.NewDiscoveryScaleKindResolver(kubeClient.Discovery())
+	scaleClient, err := scale.NewForConfig(cfg, restMapper, dynamic.LegacyAPIPathResolverFunc, scaleKindResolver)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create scale client: %w", err)
+	}
+
+	return &hpaTargetScaler{
+		mapper:      restMapper,
+		scaleClient: scaleClient,
+	}, nil
+}
+
+// Scale scales the target resource of the given HPA to the desired number of
+// replicas.
+func (s *hpaTargetScaler) Scale(ctx context.Context, hpa *autoscalingv2.HorizontalPodAutoscaler, replicas int32) error {
+	reference := fmt.Sprintf("%s/%s/%s", hpa.Spec.ScaleTargetRef.Kind, hpa.Namespace, hpa.Spec.ScaleTargetRef.Name)
+
+	targetGV, err := schema.ParseGroupVersion(hpa.Spec.ScaleTargetRef.APIVersion)
+	if err != nil {
+		return fmt.Errorf("invalid API version '%s' in scale target reference: %w", hpa.Spec.ScaleTargetRef.APIVersion, err)
+	}
+
+	targetGK := schema.GroupKind{
+		Group: targetGV.Group,
+		Kind:  hpa.Spec.ScaleTargetRef.Kind,
+	}
+
+	mappings, err := s.mapper.RESTMappings(targetGK)
+	if err != nil {
+		return fmt.Errorf("unable to determine resource for scale target reference: %w", err)
+	}
+
+	scale, targetGR, err := s.scaleForResourceMappings(ctx, hpa.Namespace, hpa.Spec.ScaleTargetRef.Name, mappings)
+	if err != nil {
+		return fmt.Errorf("failed to get scale subresource for %s: %w", reference, err)
+	}
+
+	scale.Spec.Replicas = replicas
+	_, err = s.scaleClient.Scales(hpa.Namespace).Update(ctx, targetGR, scale, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to rescale %s: %w", reference, err)
+	}
+
+	return nil
+}
+
+// scaleForResourceMappings attempts to fetch the scale for the
+// resource with the given name and namespace, trying each RESTMapping
+// in turn until a working one is found.  If none work, the first error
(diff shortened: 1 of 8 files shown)


pkg/controller/scheduledscaling/scheduled_scaling.go
```diff
@@ -183,30 +191,208 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+	// TODO: event recorder
+	err = c.adjustScaling(ctx, schedules, clusterschedules)
+	if err != nil {
+		return fmt.Errorf("failed to adjust scaling: %w", err)
+	}
+
+	return nil
+}
+
+func (c *Controller) adjustScaling(ctx context.Context, schedules []*v1.ScalingSchedule, clusterschedules []*v1.ClusterScalingSchedule) error {
```
demonCoder95, line 230: It would really help to split this in 2-3 functions to make it easier to review it.

mikkeloscar: I have split it up, hope it helps.

demonCoder95: Thanks! I'll give it a read again 🙏

pkg/controller/scheduledscaling/scaler.go
```diff
@@ -0,0 +1,118 @@
...
+
+// TargetScaler is an interface for scaling a target referenced resource in an
+// HPA to the desired replicas.
+type TargetScaler interface {
+	Scale(ctx context.Context, hpa *autoscalingv2.HorizontalPodAutoscaler, replicas int32) error
+}
+
+type hpaTargetScaler struct {
+	mapper          apimeta.RESTMapper
+	scaleNamespacer scaleclient.ScalesGetter
```
demonCoder95, line 31: The name `scaleNamespacer` is a bit confusing for me. From the code it seems to be a "typed" scale getter? Is there a way to name the variable that makes it more readable?

pkg/controller/scheduledscaling/scaler.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// NewHPATargetScaler creates a new TargetScaler that can scale resources
+// targeted by HPAs. It takes a Kubernetes client and a REST config and uses a
+// restmapper to resolve the target reference api.
```
demonCoder95, line 36: nit: `API`.

pkg/controller/scheduledscaling/scheduled_scaling.go
```diff
@@ -183,30 +191,220 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+		}
+
+		maxValue := int64(0)
+		for _, activeSchedule := range activeSchedules {
+			if activeSchedule.Value > maxValue {
+				maxValue = activeSchedule.Value
+			}
+		}
+		currentActiveSchedules[schedule.Namespace+"/"+schedule.Name] = maxValue
+	}
```
demonCoder95, line 253: Since the function `c.activeSchedules` expects a `v1.Schedule` type that's the same for both ScalingSchedule and ClusterScalingSchedule, can't we avoid the code duplication here easily? 

Maybe, having a function that takes `[]v1.Schedule` type as input and return the max value of the active schedule in there? Kind of how we detect the `activeSchedules`.

pkg/controller/scheduledscaling/scheduled_scaling.go
```diff
@@ -183,30 +191,220 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+		case "ClusterScalingSchedule":
+			value = activeSchedules[scheduleName]
+		}
+
+		expected := int64(math.Ceil(float64(value) / float64(target)))
+		if expected > highestExpected {
+			highestExpected = expected
+			highestObject = metric.Object.DescribedObject
+		}
+	}
```
demonCoder95, line 320: Let's break this out into a function like `getHighestExpected` or something similar. Seems to be an independent piece of code we can refactor easily.

pkg/controller/scheduledscaling/scheduled_scaling.go
```diff
@@ -183,30 +191,220 @@ func (c *Controller) runOnce(ctx context.Context) error {
...
+		scheduleRef := highestObject.Name
+		if highestObject.Kind == "ScalingSchedule" {
+			scheduleRef = hpa.Namespace + "/" + scheduleRef
+		}
+
+		c.recorder.Eventf(
+			hpa,
+			corev1.EventTypeNormal,
+			"ScalingAdjusted",
+			"Adjusted replicas %d -> %d based on schedule scaling metric: %s '%s'",
```
demonCoder95, line 346: Maybe it makes it more readable if we prepend the schedule name to the event message like:

```
Scaling schedule '%s' adjusted replicas %d -> %d based on metric: %s
```

Or something similar, wdyt?

pkg/server/start.go
```diff
@@ -138,6 +138,7 @@ func NewCommandStartAdapterServer(stopCh <-chan struct{}) *cobra.Command {
 	flags.DurationVar(&o.DefaultScheduledScalingWindow, "scaling-schedule-default-scaling-window", 10*time.Minute, "Default rampup and rampdown window duration for ScalingSchedules")
 	flags.IntVar(&o.RampSteps, "scaling-schedule-ramp-steps", 10, "Number of steps used to rampup and rampdown ScalingSchedules. It's used to guarantee won't avoid reaching the max scaling due to the 10% minimum change rule.")
 	flags.StringVar(&o.DefaultTimeZone, "scaling-schedule-default-time-zone", "Europe/Berlin", "Default time zone to use for ScalingSchedules.")
+	flags.Float64Var(&o.HorizontalPodAutoscalerTolerance, "horizontal-pod-autoscaler-tolerance", 0.1, "The HPA tolerance also configured in the HPA controller.")
```
demonCoder95, line 141: I see we already have the `horizontal_pod_autoscaler_tolerance` [exposed as a config-item](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/config-defaults.yaml#L510). Let's remember to set this parameter to the same value when we rollout, just to make sure they're both equal.

pkg/controller/scheduledscaling/scheduled_scaling_test.go
```diff
@@ -297,3 +306,142 @@ func checkSchedules(t *testing.T, client zalandov1.ZalandoV1Interface, schedules
...
+		{
+			msg:             "current less than 10%% below desired",
+			currentReplicas: 95, // 5.3% increase to desired
+			desiredReplicas: 100,
+			targetValue:     10, // 1000/10 = 100
+		},
+		{
+			msg:             "current more than 10%% below desired, no adjustment",
+			currentReplicas: 90, // 11% increase to desired
+			desiredReplicas: 90,
```
zaklawrencea, line 350: Shouldn't this be `100`?

mikkeloscar: Maybe I can rename something to make it more clear. But the desired, from the perspective of the test case, is to make no adjustment because >10% we don't touch, but let the HPA deal with.

pkg/controller/scheduledscaling/scheduled_scaling.go
mikkeloscar: Moved to a separate function.

pkg/controller/scheduledscaling/scheduled_scaling.go
mikkeloscar: Changed

pkg/controller/scheduledscaling/scheduled_scaling_test.go
zaklawrencea: Ah, got it! Within the scope of these tests the naming make sense, I was thinking too broad. I think we can leave it as is.

pkg/controller/scheduledscaling/scheduled_scaling.go
mikkeloscar: The problem is we have a list of `[]*v1.ScalingSchedule` and `[]*v1.ClusterScalingSchedule` which we need to unpack into `[]v1.Schedule`. Since go can't understand that different types with the same field name/type (in this case `.Spec`) could be handled the same we at least need to run through the `[]*v1.ScalingSchedule` and `[]*v1.ClusterScalingSchedule` once to unpack it.

I have tried to create an interface to handle scalingSchedule and ClusterScalingSchedule in a cosistent way. Please check if that is more readable now.

demonCoder95: Looks great now 🙂 Thanks!

pkg/server/start.go
mikkeloscar: Yes, this is the intention :)

demonCoder95:
👍

mikkeloscar:
:+1:

PR #782, opened 2025-01-29
Fix scaling adjustment edge case

mikkeloscar:
Under some conditions does the scaling adjuster, introduced in #765, not correctly up-scale resources.

An example to showcase the problem:

Scheduled scaling with target: `10000`
HPA with averageTarget: `333`
currentReplicas: `28`

In this case it's expected that it would scale to `10000 / 333 ~= 31`, however, because of the calculation in the adjustment logic is it assumed that the change is above the 10% tolerance and that the HPA will scale. This doesn't happen because the HPA assumes it's below the 10% tolerance.

The adjuster uses the following logic to calculate the _change_:

```
(expectedReplicas - currentReplicas) / currentReplicas

# (31 / 28) / 28 ~= 10.7%
```

And the HPA uses the following logic:

```
usageRatio := scheduleTarget / (hpaTarget * currentReplicas)
change = math.Abs(1.0-usageRatio)

# 10000 / (333 * 28) ~= 1.0725
# abs(1.0-1.0725) = 7.25%
```

This PR adopts the logic from the HPA such that the adjuster and HPA will come to the same conclusion and not assume the other component scales.

It adds a test case for the outlined example to validate it works.


pkg/controller/scheduledscaling/scheduled_scaling.go
```diff
@@ -266,13 +266,15 @@ func (c *Controller) adjustHPAScaling(ctx context.Context, hpa *autoscalingv2.Ho
 		return nil
 	}
 
-	highestExpected, highestObject := highestActiveSchedule(hpa, activeSchedules)
+	highestExpected, usageRatio, highestObject := highestActiveSchedule(hpa, activeSchedules, current)
 
 	highestExpected = int64(math.Min(float64(highestExpected), float64(hpa.Spec.MaxReplicas)))
 
+	fmt.Println(highestExpected, current)
```
zaklawrencea, line 273: Lets remove this print, or change it to logging.

mikkeloscar: Thanks! Fixed!

zaklawrencea:
:+1:

mikkeloscar:
:+1:

PR #785, opened 2025-02-06
Fix the check for HPA changed

### LINKED ISSUE #774: "Removing previously scheduled metrics collector" happens after every "Collected new external metric"
## Expected Behavior
Should the log "Removing previously scheduled metrics collector" happen every time kube-metrics-adapter logs "Collected new external metric"?  It seems like it should not be removing the metrics collector each time?

## Actual Behavior
"Removing previously scheduled metrics collector" happens after every "Collected new external metric"

  1. use external http collector which hits a  custom metrics api.  the metrics api returns a payload of { custom_metric: 90 } and the hpa is configured to scale based on that value. Note: the scaling is working for me.
  

## Specifications

  - Version: Latest version of kube-metrics-adapter
  - Platform: Running Kubernetes 1.29 in Amazon EKS.

mikkeloscar:
The metrics-adapter continuously checks for changes between the cached HPAs and HPAs in the cluster to detect potential changes.

The logic for checking if a cached HPA is up-to-date is broken because Metadata of an HPA, e.g. managedFields can change without the actual HPA spec changes which would require a change on the metric collector.

Fix this by only comparing HPA annotations and Spec as those are the only things that matter for comparing HPAs and deciding if the collector should be re-initialized.

Additionally change some logging to be debug logging to reduce noise if nothing changes for a while.

Fix #774

### DIFF
--- a/pkg/provider/hpa.go
+++ b/pkg/provider/hpa.go
@@ -25,6 +25,10 @@ import (
 	"github.com/zalando-incubator/kube-metrics-adapter/pkg/recorder"
 )
 
+const (
+	kubectlLastAppliedAnnotation = "kubectl.kubernetes.io/last-applied-configuration"
+)
+
 var (
 	// CollectionSuccesses is the total number of successful collections.
 	CollectionSuccesses = promauto.NewCounter(prometheus.CounterOpts{
@@ -120,7 +124,7 @@ func (p *HPAProvider) Run(ctx context.Context) {
 // updateHPAs discovers all HPA resources and sets up metric collectors for new
 // HPAs.
 func (p *HPAProvider) updateHPAs() error {
-	p.logger.Info("Looking for HPAs")
+	p.logger.Debug("Looking for HPAs")
 
 	hpas, err := p.client.AutoscalingV2().HorizontalPodAutoscalers(metav1.NamespaceAll).List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

pkg/provider/hpa.go
```diff
@@ -197,20 +201,51 @@ func (p *HPAProvider) updateHPAs() error {
...
 // equalHPA returns true if two HPAs are identical (apart from their status).
 func equalHPA(a, b autoscalingv2.HorizontalPodAutoscaler) bool {
-	// reset resource version to not compare it since this will change
-	// whenever the status of the object is updated. We only want to
-	// compare the metadata and the spec.
-	a.ObjectMeta.ResourceVersion = ""
-	b.ObjectMeta.ResourceVersion = ""
-	return reflect.DeepEqual(a.ObjectMeta, b.ObjectMeta) && reflect.DeepEqual(a.Spec, b.Spec)
+	return annotationsUpToDate(a.ObjectMeta, b.ObjectMeta, kubectlLastAppliedAnnotation) && reflect.DeepEqual(a.Spec, b.Spec)
+}
```
demonCoder95, line 218: Do we want a test-case for this? Or are the current test-cases enough?

mikkeloscar: Added test case!

demonCoder95:
👍

mikkeloscar:
:+1:

PR #795, opened 2025-03-27
Add support for filtering nakadi subscriptions

mikkeloscar:
This extends the nakadi collector to support filtering subscriptions based on fields: `owning-application,event-types,consumer-groups`

```yaml
metrics:
  - type: External
    external:
      metric:
          name: my-nakadi-consumer
          selector:
            matchLabels:
              type: nakadi
              owning-application: my-app
              event-types: event_type_v1,event_type_v2
              consumer-group: abcd
              metric-type: "consumer-lag-seconds|unconsumed-events"
```

This allows targeting a subscription without knowing the subscription-id.

### TODO

* [x] test new logic
* [x] Enhance docs in README
* [x] What fields are required when filtering? all or only some of them, adjust validation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -751,12 +751,12 @@ spec:
   - type: External
     external:
       metric:
-          name: my-nakadi-consumer
-          selector:
-            matchLabels:
-              type: nakadi
-              subscription-id: "708095f6-cece-4d02-840e-ee488d710b29"
-              metric-type: "consumer-lag-seconds|unconsumed-events"
+        name: my-nakadi-consumer
+        selector:
+          matchLabels:
+            type: nakadi
+            subscription-id: "708095f6-cece-4d02-840e-ee488d710b29"
+            metric-type: "consumer-lag-seconds|unconsumed-events"
       target:
         # value is compatible with the consumer-lag-seconds metric type.
         # It describes the amount of consumer lag in seconds before scaling
@@ -805,6 +805,27 @@ with more consumers.
 For this case you should also account for the average time for processing an
 event when defining the target.
 
+Alternative to defining `subscription-id` you can also filter based on
+`owning_application`, `event-types` and `consumer-group`:
+
+```yaml
+metrics:
+- type: External
+  external:
+    metric:
+      name: my-nakadi-consumer
+      selector:
+        matchLabels:
+          type: nakadi
+          owning-application: "example-app"
+          # comma separated list of event types
+          event-types: "example-event-type,example-event-type2"
+          consumer-group: "abcd1234"
+          metric-type: "consumer-lag-seconds|unconsumed-events"
+```
+
+This is useful in dynamic environments where the subscription ID might not be
+known at deployment time.
 
 ## HTTP Collector
 
--- a/README.md
+++ b/README.md
@@ -825,7 +825,7 @@ metrics:
 ```
 
 This is useful in dynamic environments where the subscription ID might not be
-known at deployment time.
+known before deployment time (e.g. because it's created by the same deployment).
 
 ## HTTP Collector
 
--- a/pkg/collector/nakadi_collector.go
+++ b/pkg/collector/nakadi_collector.go
@@ -93,8 +93,8 @@ func NewNakadiCollector(_ context.Context, nakadiClient nakadi.Nakadi, hpa *auto
 		subscriptionFilter.ConsumerGroup = consumerGroup
 	}
 
-	if subscriptionFilter.SubscriptionID == "" && (subscriptionFilter.OwningApplication == "" && len(subscriptionFilter.EventTypes) == 0 && subscriptionFilter.ConsumerGroup == "") {
-		return nil, fmt.Errorf("either subscription-id or one of [owning-application, event-types, consumer-group] must be specified on the metric")
+	if subscriptionFilter.SubscriptionID == "" && (subscriptionFilter.OwningApplication == "" || len(subscriptionFilter.EventTypes) == 0 || subscriptionFilter.ConsumerGroup == "") {
+		return nil, fmt.Errorf("either subscription-id or all of [owning-application, event-types, consumer-group] must be specified on the metric")
 	}
 
 	return &NakadiCollector{
--- a/pkg/collector/nakadi_collector.go
+++ b/pkg/collector/nakadi_collector.go
@@ -73,7 +73,8 @@ func NewNakadiCollector(_ context.Context, nakadiClient nakadi.Nakadi, hpa *auto
 
 	// Either subscription-id or filtering via owning-application,
 	// event-types, and consumer-group is supported. If all are defined
-	// then subscription-id is used.
+	// then only subscription-id is used and the rest of the fields are
+	// ignored.
 	subscriptionFilter := &nakadi.SubscriptionFilter{}
 	if subscriptionID, ok := config.Config[nakadiSubscriptionIDKey]; ok {
 		subscriptionFilter.SubscriptionID = subscriptionID
@@ -93,7 +94,7 @@ func NewNakadiCollector(_ context.Context, nakadiClient nakadi.Nakadi, hpa *auto
 	}
 
 	if subscriptionFilter.SubscriptionID == "" && (subscriptionFilter.OwningApplication == "" && len(subscriptionFilter.EventTypes) == 0 && subscriptionFilter.ConsumerGroup == "") {
-		return nil, fmt.Errorf("subscription-id or owning-application, event-types, and consumer-group must be specified on the metric")
+		return nil, fmt.Errorf("either subscription-id or one of [owning-application, event-types, consumer-group] must be specified on the metric")
 	}
 
 	return &NakadiCollector{
--- a/pkg/collector/nakadi_collector.go
+++ b/pkg/collector/nakadi_collector.go
@@ -3,6 +3,7 @@ package collector
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/zalando-incubator/kube-metrics-adapter/pkg/nakadi"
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -805,6 +805,27 @@ with more consumers.
...
+          type: nakadi
+          owning-application: "example-app"
+          # comma separated list of event types
+          event-types: "example-event-type,example-event-type2"
+          consumer-group: "abcd1234"
+          metric-type: "consumer-lag-seconds|unconsumed-events"
+```
+
+This is useful in dynamic environments where the subscription ID might not be
+known at deployment time.
```
ePaul, line 828:
```suggestion
This is useful in dynamic environments where the subscription ID might not be
known before deployment time (e.g. because it's created by the same deployment).
```

mikkeloscar:
:+1:

ePaul:
:+1: for the API, but I don't know enough Go to be able to judge the implementation.

pkg/collector/nakadi_collector.go
```diff
@@ -72,14 +71,39 @@ func NewNakadiCollector(_ context.Context, nakadi nakadi.Nakadi, hpa *autoscalin
 		return nil, fmt.Errorf("metric-type must be either '%s' or '%s', was '%s'", nakadiMetricTypeConsumerLagSeconds, nakadiMetricTypeUnconsumedEvents, metricType)
 	}
 
+	// Either subscription-id or filtering via owning-application,
+	// event-types, and consumer-group is supported. If all are defined
+	// then subscription-id is used.
```
demonCoder95, line 76: "If all are defined then subscription-id is used". Can you elaborate on this? I'm assuming if we already have the `susbscription-id` defined then the rest of the parameters will not be used, so we can ignore them, I guess?

pkg/collector/nakadi_collector.go
```diff
@@ -72,14 +71,39 @@ func NewNakadiCollector(_ context.Context, nakadi nakadi.Nakadi, hpa *autoscalin
...
+		eventTypes := strings.Split(nakadiEventTypes, ",")
+		subscriptionFilter.EventTypes = eventTypes
+	}
+
+	if consumerGroup, ok := config.Config[nakadiConsumerGroupKey]; ok {
+		subscriptionFilter.ConsumerGroup = consumerGroup
+	}
+
+	if subscriptionFilter.SubscriptionID == "" && (subscriptionFilter.OwningApplication == "" && len(subscriptionFilter.EventTypes) == 0 && subscriptionFilter.ConsumerGroup == "") {
+		return nil, fmt.Errorf("subscription-id or owning-application, event-types, and consumer-group must be specified on the metric")
```
demonCoder95, line 96: I think the error message can be improved a little with something like 
```golang
"either subscription-id or one of [owning-application, event-types, consumer-group] must be specified on the metric"
```
Just to make it clearer to the user what we expect from them.
Just to clar

pkg/provider/metric_store.go
```diff
@@ -140,7 +140,7 @@ func (s *MetricStore) insertCustomMetric(value custom_metrics.MetricValue) {
 		labelsKey = hashLabelMap(selector.MatchLabels)
 	}
 
-	metric := metricName(value.Metric.Name)
+	metric := metricName(strings.ToLower(value.Metric.Name))
```
demonCoder95, line 143: Was there a particular reason to introduce this change? Asking out of curiousity.

pkg/nakadi/nakadi.go
```diff
@@ -62,6 +62,90 @@ func (c *Client) UnconsumedEvents(ctx context.Context, subscriptionID string) (i
...
+	}
+	defer resp.Body.Close()
+
+	d, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("[nakadi subscriptions] unexpected response code: %d (%s)", resp.StatusCode, string(d))
```
demonCoder95, line 115: I like the prefix of `[nakadi subsriptions]` prefix in the error string. We should add it to the rest of the error messages in the function as well.

pkg/nakadi/nakadi.go
```diff
@@ -62,6 +62,90 @@ func (c *Client) UnconsumedEvents(ctx context.Context, subscriptionID string) (i
...
 }
 
+type SubscriptionFilter struct {
+	SubscriptionID    string
+	OwningApplication string
+	EventTypes        []string
+	ConsumerGroup     string
+}
+
+func (c *Client) subscriptions(ctx context.Context, filter *SubscriptionFilter, href string) ([]string, error) {
```
demonCoder95, line 72: Just a general thought, this raw HTTP stuff doesn't look very nice. I would imagine there are many use cases for doing this kind of a thing with the nakadi API:
1. Have some query parameters.
2. Send a query requests to nakadi endpoint.
3. Process the response

Don't we have some kind of a client library we can use? 🤔

pkg/nakadi/nakadi.go
```diff
@@ -83,42 +167,63 @@ type statsPartition struct {
...
+			return nil, err
+		}
 
-	var result statsResp
-	err = json.Unmarshal(d, &result)
-	if err != nil {
-		return nil, err
-	}
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("[nakadi stats] unexpected response code: %d (%s)", resp.StatusCode, string(d))
```
demonCoder95, line 212: Same comment here, we should prepend the other error messages with `[nakadi stats]` prefix as well, I think.

pkg/nakadi/nakadi.go
demonCoder95: Same question about the `stats` endpoint, there's a lot of raw HTTP stuff there.

pkg/nakadi/nakadi.go
```diff
@@ -83,42 +167,63 @@ type statsPartition struct {
 // stats returns the Nakadi stats for a given subscription ID.
```
demonCoder95, line 167: We have to update this comment here

pkg/collector/nakadi_collector.go
mikkeloscar: Yes, the other fields, except `subscription-id` are used for find a `subscription-id`. If `subscription-id` is already defined the other fields are not used for filtering.

I have updated the comment a bit to clarify this.

pkg/nakadi/nakadi.go
mikkeloscar: added

pkg/nakadi/nakadi.go
mikkeloscar: Added

pkg/provider/metric_store.go
mikkeloscar: This was intended for another bug fix unrelated to this, I accidentally commit this.

Removed again.

pkg/collector/nakadi_collector.go
mikkeloscar: Updated

pkg/collector/nakadi_collector.go
```diff
@@ -72,14 +71,40 @@ func NewNakadiCollector(_ context.Context, nakadi nakadi.Nakadi, hpa *autoscalin
...
+		subscriptionFilter.EventTypes = eventTypes
+	}
+
+	if consumerGroup, ok := config.Config[nakadiConsumerGroupKey]; ok {
+		subscriptionFilter.ConsumerGroup = consumerGroup
+	}
+
+	if subscriptionFilter.SubscriptionID == "" && (subscriptionFilter.OwningApplication == "" && len(subscriptionFilter.EventTypes) == 0 && subscriptionFilter.ConsumerGroup == "") {
+		return nil, fmt.Errorf("either subscription-id or one of [owning-application, event-types, consumer-group] must be specified on the metric")
+	}
```
ePaul, line 98: We actually need the whole triple of [owning-application, event-types, consumer-group] to determine the subscription ID uniquely. (While consumer-group is optional, it is nowadays considered good form to always have it, and I guess we can accept if this doesn't support consumer-group-less subscriptions.)

```suggestion
	if subscriptionFilter.SubscriptionID == "" && (subscriptionFilter.OwningApplication == "" || len(subscriptionFilter.EventTypes) == 0 || subscriptionFilter.ConsumerGroup == "") {
		return nil, fmt.Errorf("either subscription-id or all of [owning-application, event-types, consumer-group] must be specified on the metric")
	}
```

pkg/nakadi/nakadi.go
ePaul: There is https://github.bus.zalan.do/aruha/nakadi-go, but that's a Zalando-internal library, not supported, and (from a short look) only has the "get subscription" without the consumer group parameter.

pkg/nakadi/nakadi_test.go
```diff
@@ -52,9 +76,10 @@ func TestQuery(tt *testing.T) {
 			consumerLagSeconds: 2,
 		},
 		{
-			msg:    "test getting multiple event-types",
-			status: http.StatusOK,
-			responseBody: `{
+			msg:                "test getting multiple event-types",
+			status:             http.StatusOK,
+			subscriptionFilter: &SubscriptionFilter{SubscriptionID: "id"},
```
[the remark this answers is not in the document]
mikkeloscar: This test you comment on was just adopted to support the new filter, but similar to before where the only option was subscription ID it only passes that in the filter.

It's testing handling a response with multiple event types.

Since we mock the nakadi stats API server we can't really meaningfully test if supplying multiple event types in the filter would work as. We would need an integration test with Nakadi stats API to test this. I think we have to live without that for now as that would require quite some effort to setup.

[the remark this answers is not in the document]
demonCoder95: I see. That makes sense then. Thanks!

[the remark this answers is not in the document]
demonCoder95: I see. That makes sense then. Thanks!

pkg/collector/nakadi_collector.go
demonCoder95: Thanks!

mikkeloscar:
:+1:

PR #799, opened 2025-04-09
Add short names for (cluster)scalingschedule CRDs

zaklawrencea:
Add short names for the `ClusterScalingSchedule` and `ScalingSchedule` CRDs. Makes life a little easier by not needing to type out the long CRD names every time.


zaklawrencea:
:+1:

docs/cluster_scaling_schedules_crd.yaml
```diff
@@ -12,6 +12,10 @@ spec:
     listKind: ClusterScalingScheduleList
     plural: clusterscalingschedules
     singular: clusterscalingschedule
+    shortNames:
+      - css
+      - clustersched
+      - clusterschedule
```
mikkeloscar, line 18: These files are generated, the change needs to be done in https://github.com/zalando-incubator/kube-metrics-adapter/blob/master/pkg/apis/zalando.org/v1/types.go (There is some annotation to generate this)

zaklawrencea: Oh you're right, let me take another look at this.

zaklawrencea: Updated. It looks like the `all` category was previously missing from the `cluster_scaling_schedules` as there were 2 `kubebuilder:resource` annotations defined with only the final one actually taking effect (`scope`).

mikkeloscar:
:+1:

zaklawrencea:
:+1:

PR #802, opened 2025-04-28
Remove all category from ClusterScalingSchedule

zaklawrencea:
Same as https://github.com/zalando-incubator/kubernetes-on-aws/pull/9302

> ClusterScalingSchedule is the only global CRD that is displayed as part of kubectl get all. This is fairly confusing when you want to see "all" resources in a namespace (knowing that all is not really everything) but you constantly also see the ClusterScalingSchedule everytime.

> This was recently introduced by a bug fix but turns out the old (bugged) behaviour was actually better.

### DIFF
--- a/docs/cluster_scaling_schedules_crd.yaml
+++ b/docs/cluster_scaling_schedules_crd.yaml
@@ -8,8 +8,6 @@ metadata:
 spec:
   group: zalando.org
   names:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

zaklawrencea:
:+1:

Issue #804, opened 2025-05-13
Latest version 0.2.4 not found


sarasensible:
Similar to https://github.com/zalando-incubator/kube-metrics-adapter/issues/731 except with version 0.2.4.  It has been two months however and still no Docker image for the newest release.

docker pull registry.opensource.zalan.do/teapot/kube-metrics-adapter:v0.2.4

Fails with
Error response from daemon: manifest for registry.opensource.zalan.do/teapot/kube-metrics-adapter:v0.2.4 not found: manifest unknown: manifest unknown

I recommend investing in a CI/CD pipeline to ensure that new docker images and possible new versions of the helm chart are released in tandem with new code.

sarasensible:
Is this repository abandoned?

szuecs:
@sarasensible no, we are using it in production and many other orgs, too.

sarasensible:
Awesome!  The following command works:
`ghcr.io/zalando-incubator/kube-metrics-adapter:v0.2.4`

Thanks so much!

PR #805, opened 2025-05-16
helm: fix wrong image as we moved to ghcr

szuecs:
fix: https://github.com/zalando-incubator/kube-metrics-adapter/issues/804


szuecs:
@sarasensible I grepped the image url to check where we have it and I found only this one.

PR #839, opened 2025-10-29
Run code gen via Go 1.24 go tool

mikkeloscar:
Updates the code generation to use `go tool` instead of a custom `tools.go`. This is a new feature since Go 1.24


katyanna:
👍

PR #855, opened 2025-12-18
Enrich GetMetrics errors with metric identifier

mikkeloscar:
If an error is returned from `GetMetrics()` it's not obvious which metric the error is relevant for.

This enhances the error to include a formatted string of the type name identifier for the metric such that it's clear for which metric it failed.


demonCoder95:
👍

PR #862, opened 2026-02-06
retry failed nakadi requests and respect rate limits

bedakb:
# One-line summary

Properly retry failed Nakadi requests and respect the rate-limits recommendation by relying on Retry-After header on 429,

## Description
This PR introduces a retry mechanism for failed Nakadi requests using an exponential backoff algorithm. Additionally, when a 429 status code is returned with a Retry-After header, the system will respect the specified wait time before retrying.

This PR will also help handle transient network errors. Previously, operations would simply fail and only retry when the next collection interval was triggered.

## Types of Changes

- Refactor/improvements

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes

### DIFF
--- a/pkg/nakadi/nakadi_test.go
+++ b/pkg/nakadi/nakadi_test.go
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
+	"github.com/cenkalti/backoff/v5"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/nakadi/nakadi_test.go
```diff
@@ -210,11 +212,70 @@ func TestQuery(tt *testing.T) {
...
+}
+
+func assertErrorMessage(t *testing.T, expected, actual error) {
+	t.Helper()
+	if expected != nil {
+		assert.EqualError(t, actual, expected.Error())
+	} else {
+		assert.NoError(t, actual)
+	}
+}
```
demonCoder95: Is this function really needed? I see it's used only for assertions in lines 215 and 218 and we only use the `assert.EqualError` branch of it. Did you have something else in mind? Otherwise, let's keep the assertions as they are, and we can drop this extra function. Just replace the `assert.Equal` with `assert.EqualError` in the original code, instead.

PR #868, opened 2026-02-18
fix: Only format if there is actual an error

Shinzu:
# One-line summary

only format the error if there is actual an error getting the metrics

## Description

currently on every metric collection an error is reported since Error field is not nil 

this floods the logs with something like:

```
time="2026-02-18T15:44:31Z" level=error msg="Failed to collect metrics: getting metrics for External/my-app {type=nakadi,metric-type=unconsumed-events,subscription-id=xx-xx} failed: %!w(<nil>)" provider=hpa
```

so we format the error only if the metrics collections return an error

## Types of Changes

## Tasks

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes

### DIFF
--- a/pkg/provider/hpa.go
+++ b/pkg/provider/hpa.go
@@ -384,14 +384,13 @@ func (t *CollectorScheduler) Add(resourceRef resourceReference, typeName collect
 func collectorRunner(ctx context.Context, typeName collector.MetricTypeName, collector collector.Collector, metricsc chan<- metricCollection) {
 	for {
 		values, err := collector.GetMetrics(ctx)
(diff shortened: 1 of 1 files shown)


pkg/provider/hpa.go
```diff
@@ -385,9 +385,13 @@ func collectorRunner(ctx context.Context, typeName collector.MetricTypeName, col
 	for {
 		values, err := collector.GetMetrics(ctx)
```
linki:
```suggestion
```

PR #869, opened 2026-02-19
fix: Nil check Metric.Selector in MetricTypeName

Shinzu:
# One-line summary

Nil check Metric.Selector before try to get len()

## Description

Metric.Selector can be nil and we need to check this before we try to get the len of the MatchLabels other wise we get an error message like:

```
time="2026-02-19T11:56:57Z" level=error msg="Failed to collect metrics: getting metrics for %!s(PANIC=String method: runtime error: invalid memory address or nil pointer dereference) failed: %!w(<nil>)" provider=hpa
```

This prevents panic during error formatting, revealing the actual underlying errors instead of masking them.

## Types of Changes

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes

### DIFF
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -185,7 +185,7 @@ type MetricTypeName struct {
 
 func (m MetricTypeName) String() string {
 	str := fmt.Sprintf("%s/%s", m.Type, m.Metric.Name)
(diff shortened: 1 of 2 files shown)


linki:
👍

